PR #1, opened 2023-12-04
Moved the entire framework to the shepherd repo.

igomez06:
Updated dependencies, and moved packages that were in rancher/rancher into shepherd that are needed in the framework.


git-ival (approved):
.gitignore should be added with relevant files that should be excluded
  - `.DS_Store` files should be part of the ignored files

Overall, the structure looks good although I'm not very familiar with how the codegen works. It seems like everything should work out-of-the-box though

.DS_Store
git-ival, line 1: Remove all .DS_Store files

git-ival (changes requested):
Add .gitignore file with files that should be excluded from VCS

git-ival:
Also maybe this is a future PR type of thing, but we will probably want some guardrails on PRs (requiring X # of approvals, etc.) As well as some documentation on branching strategies and when `go.mod` can be updated, etc.

.DS_Store
igomez06: Got it on it now.

igomez06:
> Also maybe this is a future PR type of thing, but we will probably want some guardrails on PRs (requiring X # of approvals, etc.) As well as some documentation on branching strategies and when `go.mod` can be updated, etc.

As of right now, we're not planning on branching the framework based on release line. All the updated extensions will be in main; I might be wrong, but I don't foresee us needing to do that. As for your other question, are you referring to the go.mod file here? Or the one in rancher/rancher?

igomez06:
> > As of right now, we're not planning on branching the framework based on release line. All the updated extensions will be in main; I might be wrong, but I don't foresee us needing to do that. As for your other question, are you referring to the go.mod file here? Or the one in rancher/rancher?
> 
> I was referring to the go.mod in this repo.

We'll try and keep it consistent to the latest version of rancher, and go from there. I think this will be up to us to police, but will definitely document that the dependencies should resemble what is going on in rancher/rancher ...

pkg/codegen/generator/generator.go
```diff
@@ -0,0 +1,64 @@
...
+// 	k8sOutputPackage := path.Join(outputDir, baseK8s, groupPath, version)
+
+// 	if err := generator.GenerateControllerForTypes(&types.APIVersion{
+// 		Version: version,
+// 		Group:   group,
+// 		Path:    fmt.Sprintf("/k8s/%s-%s", groupPath, version),
+// 	}, k8sOutputPackage, nsObjs, objs); err != nil {
+// 		panic(err)
+// 	}
+// }
```
caliskanugur, line 48: This part seems from the dev stage, would we need this part in the future?

pkg/codegen/generator/generator.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+
+	"github.com/rancher/norman/generator"
+	"github.com/rancher/norman/types"
+)
+
+var (
+	outputDir  = "./"
+	baseCattle = "clients/rancher/generated"
+	// baseK8s    = "norman"
```
caliskanugur, line 14: I've got the same nit for this one

generator.go
```diff
@@ -0,0 +1,4 @@
+//go:generate go run pkg/codegen/generator/cleanup/main.go
```
caliskanugur, line 1: I don't have a strong opinion; should this file be in the root? An alternative might be moving this to `pkg/codegen` but that might cause us to nest the current codegen dir one layer more

caliskanugur (commented):
Looks really solid to me; I've got one nit and a question

pkg/codegen/generator/generator.go
igomez06: ah you're right, this wasn't needed anymore at least for now ... Since we're going to include wrangler anyways

generator.go
igomez06: It needs to be in root, without it in root, the controllers and clients created end up in weird places.

caliskanugur (approved):
Looks really good to me, super exciting!

.gitignore
git-ival, line 1: Can we add the following to this?
```
# Binaries for programs and plugins
*.exe
*.exe~
*.dll
*.so
*.dylib

# Test binary, built with `go test -c`
*.test

# Output of the go coverage tool, specifically when used with LiteIDE
*.out

# Go workspace files
go.work
go.work.sum

# Misc
*.pem
```

git-ival (changes requested):

git-ival (approved):
LGTM

PR #2, opened 2023-12-08
Updated packages in go.mod

igomez06:
A lot of the dependency version didn't match what was in rancher/rancher, so when import shepherd into rancher/rancher it was causing unnecessary dependency upgrades. Also ported over the most recent updates.


extensions/workloads/pods/verify.go
```diff
@@ -5,11 +5,11 @@ import (
 	"time"
 
 	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
-	projectv3 "github.com/rancher/rancher/pkg/client/generated/project/v3"
```
git-ival, line 8: Whats the reason for swapping to appv1? Are these imports equivalent?

igomez06: No, one is generated for norman, and the other is the native struct we want for a v1 call. This should have never been merged like this and if it did, it should come from our own generated struct. There is so much wrong with with this import.

git-ival: Understood 😅

git-ival (approved):
LGTM

slickwarren (approved):
lgtm

markusewalker (approved):
LGTM

go.mod
```diff
@@ -13,76 +13,90 @@ replace (
...
+	github.com/rancher/rancher/pkg/apis v0.0.0-20230915232223-a9ea4ce4a5ba
+	github.com/rancher/rke v1.5.0
+	github.com/rancher/system-upgrade-controller/pkg/apis v0.0.0-20210727200656-10b094e30007
 	github.com/rancher/types v0.0.0-20220328215343-4370ff10ecd5
 	github.com/rancher/wrangler v1.1.1
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.7.0
 	github.com/stretchr/testify v1.8.4
-	golang.org/x/crypto v0.14.0
+	golang.org/x/crypto v0.12.0
```
caliskanugur, line 38: This one seems 0.14 on 2.8 r/r

caliskanugur (approved):
Looks good to me

PR #3, opened 2023-12-15
[v2.7] Updated packages in go.mod

igomez06:
A lot of the dependency version didn't match what was in rancher/rancher, so when import shepherd into rancher/rancher it was causing unnecessary dependency upgrades. Also ported over the most recent updates.


slickwarren (approved):

go.mod
```diff
@@ -3,87 +3,99 @@ module github.com/rancher/shepherd
...
-	github.com/rancher/rancher/pkg/client v0.0.0-20231127155507-649110f64f1e
-	github.com/rancher/rke v1.5.0-rc9
-	github.com/rancher/system-upgrade-controller/pkg/apis v0.0.0-20231123175309-a70667b30261
+	github.com/pkg/sftp v1.13.5
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.52.0
+	github.com/rancher/apiserver v0.0.0-20230810214006-d134152ce6fb
+	github.com/rancher/eks-operator v1.2.2
+	github.com/rancher/fleet/pkg/apis v0.0.0-20230821130910-23e1c146755a
+	github.com/rancher/gke-operator v1.1.6
+	github.com/rancher/lasso v0.0.0-20230810170912-a86ba6c749aa
```
caliskanugur, line 28: It seems like the previous lasso version `v0.0.0-20230830164424-d684fdeb6f29` on 2.7 r/r

go.mod
```diff
@@ -3,87 +3,99 @@ module github.com/rancher/shepherd
...
 	github.com/stretchr/testify v1.8.4
-	golang.org/x/crypto v0.14.0
+	golang.org/x/crypto v0.12.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.28.3
-	k8s.io/apimachinery v0.28.3
-	k8s.io/apiserver v0.28.3
-	k8s.io/cli-runtime v0.27.6
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
```
caliskanugur, line 41: It seems these two have different versions, I think it shouldn't matter because the above replaces though, right?

igomez06: These just got updated, weird that it still has the replace if they wanted an updated version

igomez06: updated anyway

caliskanugur (approved):
Looks good to me

PR #16, opened 2024-01-04
Adding extensions for rbac checks

anupama2501:
Extension in this PR are to enhance the extensions for rbac checks. Once approved, this needs a back port in 2.7 and 2.8 branches as well.

Rancher PR that uses the extensions from this PR: https://github.com/rancher/rancher/pull/43661

### DIFF
--- a/extensions/kubeapi/rbac/list.go
+++ b/extensions/kubeapi/rbac/list.go
@@ -32,7 +32,6 @@ func ListRoleBindings(client *rancher.Client, clusterName, namespace string, lis
 
 		rbList.Items = append(rbList.Items, *rb)
 	}
-
 	return rbList, nil
 }
 
@@ -55,7 +54,6 @@ func ListClusterRoleBindings(client *rancher.Client, clusterName string, listOpt
 		if err != nil {
 			return nil, err
 		}
-
 		crbList.Items = append(crbList.Items, *crb)
 	}
 
@@ -81,7 +79,6 @@ func ListGlobalRoleBindings(client *rancher.Client, listOpt metav1.ListOptions) 
 		if err != nil {
 			return nil, err
 		}
-
 		grbList.Items = append(grbList.Items, *grb)
 	}
 
@@ -107,7 +104,6 @@ func ListClusterRoleTemplateBindings(client *rancher.Client, listOpt metav1.List
 		if err != nil {
 			return nil, err
 		}
-
 		crtbList.Items = append(crtbList.Items, *crtb)
 	}
 
@@ -133,7 +129,6 @@ func ListGlobalRoles(client *rancher.Client, listOpt metav1.ListOptions) (*v3.Gl
 		if err != nil {
 			return nil, err
 		}
-
 		grList.Items = append(grList.Items, *gr)
 	}
 
@@ -159,9 +154,34 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
 		if err != nil {
 			return nil, err
 		}
-
 		rtList.Items = append(rtList.Items, *rt)
 	}
 
 	return rtList, nil
 }
+
+// ListProjectRoleTemplateBindings is a helper function that uses the dynamic client to list projectroletemplatebindings from local cluster.
+func ListProjectRoleTemplateBindings(client *rancher.Client, listOpt metav1.ListOptions) (*v3.ProjectRoleTemplateBindingList, error) {
+	dynamicClient, err := client.GetDownStreamClusterClient(LocalCluster)
+	if err != nil {
+		return nil, err
+	}
+
+	unstructuredList, err := dynamicClient.Resource(ProjectRoleTemplateBindingGroupVersionResource).Namespace("").List(context.TODO(), listOpt)
+	if err != nil {
+		return nil, err
+	}
+
+	prtbList := new(v3.ProjectRoleTemplateBindingList)
+	for _, unstructuredPRTB := range unstructuredList.Items {
+		prtb := &v3.ProjectRoleTemplateBinding{}
+		err := scheme.Scheme.Convert(&unstructuredPRTB, prtb, unstructuredPRTB.GroupVersionKind())
+		if err != nil {
+			return nil, err
+		}
+
+		prtbList.Items = append(prtbList.Items, *prtb)
+	}
+	return prtbList, nil
+}
+
--- a/extensions/kubeapi/rbac/rbac.go
+++ b/extensions/kubeapi/rbac/rbac.go
@@ -1,7 +1,13 @@
 package rbac
 
 import (
+	"context"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/pkg/api/scheme"
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
@@ -67,9 +73,31 @@ var RoleTemplateGroupVersionResource = schema.GroupVersionResource{
 	Resource: "roletemplates",
 }
 
-// ProjectRoleTemplateBindingGroupVersionResource is the required Group Version Resource for accessing project role template bindings in a cluster, using the dynamic client.
+// ProjectRoleTemplateBindingGroupVersionResource is the required Group Version Resource for accessing projectroletemplatebindings in a cluster, using the dynamic client.
 var ProjectRoleTemplateBindingGroupVersionResource = schema.GroupVersionResource{
 	Group:    GroupName,
 	Version:  Version,
 	Resource: "projectroletemplatebindings",
+}
+
+
+// GetProjectRoleTemplateBindingsByName is a helper function that uses the dynamic client to get a projectroletemplatebinding from local cluster.
+func GetProjectRoleTemplateBindingsByName(client *rancher.Client, getOpts metav1.GetOptions, name string) (*v3.ProjectRoleTemplateBinding, error) {
+	dynamicClient, err := client.GetDownStreamClusterClient(localcluster)
+	if err != nil {
+		return nil, err
+	}
+	projectRoleTemplateBindingResource := dynamicClient.Resource(ProjectRoleTemplateBindingGroupVersionResource)
+	unstructuredResp, err := projectRoleTemplateBindingResource.Get(context.TODO(),name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	prtb := &v3.ProjectRoleTemplateBinding{}
+
+	err = scheme.Scheme.Convert(unstructuredResp, prtb, unstructuredResp.GroupVersionKind())
+	if err != nil {
+		return nil, err
+	}
+	return prtb, nil
 }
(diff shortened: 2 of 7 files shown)


extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
+	if err != nil {
+		return nil, err
+	}
+
+	unstructuredList, err := dynamicClient.Resource(ProjectRoleTemplateBindingGroupVersionResource).Namespace("").List(context.TODO(), listOpt)
+	if err != nil {
+		return nil, err
+	}
+
+	crtbList := new(v3.ClusterRoleTemplateBindingList)
```
Priyashetty17, line 181: Shouldn't this be: `prtbList := new(v3.ProjectRoleTemplateBindingList)` ?

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
+	for _, unstructuredCRTB := range unstructuredList.Items {
```
Priyashetty17, line 182: Can the references to crtb in this function be changed to prtb?

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
+		crtb := &v3.ClusterRoleTemplateBinding{}
```
Priyashetty17, line 183: Shouldn't this be: `prtb := &v3.ProjectRoleTemplateBinding{}` ?

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
 
 	return rtList, nil
 }
+
+// ListProjectRoleTemplateBindings is a helper function that uses the dynamic client to list projectroletemplatebindings from local cluster.
+func ListProjectRoleTemplateBindings(client *rancher.Client, listOpt metav1.ListOptions) (*v3.ClusterRoleTemplateBindingList, error) {
```
Priyashetty17, line 170: Shouldn't this function return `v3.ProjectRoleTemplateBindingList` instead of `v3.ClusterRoleTemplateBindingList`?

Priyashetty17 (changes requested):
Overall lgtm except for the ListProjectRoleTemplateBindings function where there are references to CRTB instead of PRTB.

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
+		return nil, err
+	}
+
+	unstructuredList, err := dynamicClient.Resource(ProjectRoleTemplateBindingGroupVersionResource).Namespace("").List(context.TODO(), listOpt)
+	if err != nil {
+		return nil, err
+	}
+
+	prtbList := new(v3.ProjectRoleTemplateBindingList)
+	for _, unstructuredCRTB := range unstructuredList.Items {
```
Priyashetty17, line 182: Can we rename `unstructuredCRTB` to `unstructuredPRTB`?

extensions/kubeapi/rbac/rbac.go
```diff
@@ -66,3 +72,32 @@ var RoleTemplateGroupVersionResource = schema.GroupVersionResource{
 	Version:  Version,
 	Resource: "roletemplates",
 }
+
+// ProjectRoleTemplateBindingGroupVersionResource is the required Group Version Resource for accessing projectroletemplatebindings in a cluster, using the dynamic client.
+var ProjectRoleTemplateBindingGroupVersionResource = schema.GroupVersionResource{
+	Group:    GroupName,
+	Version:  Version,
+	Resource: "projectroletemplatebindings",
```
Priyashetty17, line 80: Can we define this as a constant in rbac.go?

extensions/kubeapi/rbac/rbac.go
```diff
@@ -66,3 +72,32 @@ var RoleTemplateGroupVersionResource = schema.GroupVersionResource{
 	Version:  Version,
 	Resource: "roletemplates",
```
Priyashetty17, line 73: Can we define this as a constant in rbac.go?

extensions/kubeapi/rbac/rbac.go
anupama2501: I thought of that but this is used only once, we can leave this as is.

extensions/kubeapi/rbac/rbac.go
anupama2501: this is only used once in the entire package so we can leave it as is. Are you in any of your tests going to use this in the rbac suite?

Priyashetty17: I don't think I do. We can leave it as is

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
 }
+
+// ListProjectRoleTemplateBindings is a helper function that uses the dynamic client to list projectroletemplatebindings from local cluster.
+func ListProjectRoleTemplateBindings(client *rancher.Client, listOpt metav1.ListOptions) (*v3.ProjectRoleTemplateBindingList, error) {
+	dynamicClient, err := client.GetDownStreamClusterClient(localcluster)
+	if err != nil {
+		return nil, err
+	}
+
+	unstructuredList, err := dynamicClient.Resource(ProjectRoleTemplateBindingGroupVersionResource).Namespace("").List(context.TODO(), listOpt)
```
Priyashetty17, line 176: Since PRTBs are namespaced resource, should we not pass in the namespace here?

dasarinaidu (approved):
LGTM.

Priyashetty17 (approved):
lgtm

extensions/kubeapi/rbac/list.go
```diff
@@ -165,3 +165,28 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
 
 	return rtList, nil
 }
+
+// ListProjectRoleTemplateBindings is a helper function that uses the dynamic client to list projectroletemplatebindings from local cluster.
+func ListProjectRoleTemplateBindings(client *rancher.Client, listOpt metav1.ListOptions) (*v3.ProjectRoleTemplateBindingList, error) {
```
caliskanugur, line 170: I don't know if this would be in this PR's scope, but could you check other existing kubeapi lists (f.e. ingress list)? We had to do a refactor especially to grab names from a list. Some should have a method called `names`.

extensions/rbac/rbac.go
```diff
@@ -0,0 +1,18 @@
...
+	RoleProjectMember             = "project-member"
+	RoleCreateNS                  = "create-ns"
+	RoleReadOnly                  = "read-only"
+	RoleCustomManageProjectMember = "projectroletemplatebindings-manage"
+	RoleCustomCreateNS            = "create-ns"
+	ActiveStatus                  = "active"
+	ForbiddenError                = "403 Forbidden"
+	KubeConfigTokenSettingID      = "kubeconfig-default-token-ttl-minutes"
+	DefaultNamespace              = "fleet-default"
+)
```
caliskanugur, line 18: This looks good! Same as my other comment, maybe something to think about - and probably out of scope for this PR. 

Just because these are mostly strings, you might find adding a stringer method beneficial as these expand. It can also help it separate these constants further by the custom types you create, f.e.:

```
type Role string

	ClusterOwner         Role     = "cluster-owner"
	ClusterMember       Role      = "cluster-member"
	ProjectOwner          Role   = "project-owner"
	ProjectMember        Role     = "project-member"
	CreateNS                  Role  = "create-ns"
	ReadOnly                  Role  = "read-only"
	CustomManageProjectMember  Role = "projectroletemplatebindings-manage"
	CustomCreateNS        Role    = "create-ns"

func (r Role) String() string {
	return string(r)
}
```

More examples can be found in `provisioninginput` package.

While writing the above snippet, I saw that `create-ns` already has two different variables. It can prevent that too.

extensions/settings/update.go
```diff
@@ -0,0 +1,21 @@
+package settings
+
+import (
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
+)
+
+func EditGlobalSettings(steveclient *v1.Client, globalSetting *v1.SteveAPIObject, value string) (*v1.SteveAPIObject, error) {
```
caliskanugur, line 8: This one needs godoc comments. Another nit; it seems like a `UpdateGlobalSettings`, to unify with the file name.

caliskanugur (commented):
Looks good, I've got a few questions, comments and nits

extensions/kubeapi/rbac/list.go
anupama2501: I went through the names for ingressList. In that we are currently accepting the ingressList from the helper ListIngresses and returns a sorted list of names. In rbac, since we have different collections in the list file, we might need to add a generic function and then use that function sort the names. We will add a qa task for it and brainstorm on that part later for now. what do you think?

caliskanugur: Awesome, definitely agreed, I think it requires brainstorming and backlog rather than implementing it to this PR right now

extensions/rbac/rbac.go
```diff
@@ -0,0 +1,22 @@
...
+	RestrictedAdmin           Role = "restricted-admin"
+	StandardUser              Role = "user"
+	ClusterOwner              Role = "cluster-owner"
+	ClusterMember             Role = "cluster-member"
+	ProjectOwner              Role = "project-owner"
+	ProjectMember             Role = "project-member"
+	CreateNS                  Role = "create-ns"
+	ReadOnly                  Role = "read-only"
+	CustomManageProjectMember Role = "projectroletemplatebindings-manage"
+	CustomCreateNS            Role = "create-ns"
```
caliskanugur, line 15: Seems like this one and `CreateNS` have the same value

extensions/rbac/rbac.go
```diff
@@ -0,0 +1,40 @@
...
+	ActiveStatus                   = "active"
+	ForbiddenError                 = "403 Forbidden"
+	DefaultNamespace               = "fleet-default"
+)
+
+func (r Role) String() string {
+	return string(r)
+}
+
+func SetupUser(client *rancher.Client, globalRole string) (user *management.User, userClient *rancher.Client, err error) {
```
caliskanugur, line 30: Could you add godocs for these public funcs/methods?

caliskanugur (commented):
Looks good, I've only one comment

caliskanugur (approved):
Looks good to me

extensions/rbac/rbac.go
```diff
@@ -0,0 +1,41 @@
...
+	ActiveStatus                   = "active"
+	ForbiddenError                 = "403 Forbidden"
+	DefaultNamespace               = "fleet-default"
+)
+
+func (r Role) String() string {
+	return string(r)
+}
+
+// SteupUser is a helper to create a global role and a client for the user.
```
Priyashetty17, line 30: Typo: SteupUser --> should be SetupUser

anupama2501: good catch!

igomez06 (approved):
LGTM

PR #17, opened 2024-01-04
Refactor Apply function to Command and enhance job naming

brudnak:
The Apply function in 'kubectl' extension has been refactored to Command to improve clarity of function's action. It now includes the option for the function to execute an array of provided commands. The job naming mechanism was also adjusted to generate unique names using a random string for each job, improving tracking and organization of jobs.


extensions/kubectl/command.go
```diff
@@ -18,17 +19,20 @@ import (
...
 const volumeName = "config"
 
-// Apply is a helper function that call a helper to create a Job and executing kubectl apply by copying the yaml
-// in the pod. It returns the output from the pod logs.
-func Apply(client *rancher.Client, cluster *apisV1.Cluster, yamlContent *management.ImportClusterYamlInput, clusterID string) (string, error) {
+// Command creates a kubernetes job to execute the provided commands.
+// It initializes necessary volume and volume mounts for the job and overrides securityContext to execute the commands.
+// If yamlContent is provided, an init container is created to write this content to a file.
+// After creating the job it waits for the execution of the job and retrieves the logs of the executed commands
+func Command(client *rancher.Client, cluster *apisV1.Cluster, yamlContent *management.ImportClusterYamlInput, clusterID string, command []string) (string, error) {
```
anupama2501, line 26: Should we remove explain.go? because with command.go we could use it for both apply and explain if I understand it correctly?

extensions/kubectl/command.go
```diff
@@ -18,17 +19,20 @@ import (
...
 	var user int64
 	var group int64
 	imageSetting, err := client.Management.Setting.ByID(rancherShellSettingID)
 	if err != nil {
 		return "", err
 	}
 
-	jobName := JobName + "-apply"
+	id := namegen.RandStringLower(6)
```
anupama2501, line 34: any specific reason to use the integer 6 or its just random? may be we can use namegen.AppendRandomString instead?

extensions/kubectl/command.go
```diff
@@ -60,11 +64,12 @@ func Apply(client *rancher.Client, cluster *apisV1.Cluster, yamlContent *managem
...
-	initContainer := workloads.NewContainer("copy-yaml", imageSetting.Value, corev1.PullAlways, initVolumeMount, nil, initContainerCommand, nil, nil)
-	jobTemplate.Spec.Template.Spec.InitContainers = append(jobTemplate.Spec.Template.Spec.InitContainers, initContainer)
+	if yamlContent != nil {
+		initContainerCommand := []string{"sh", "-c", fmt.Sprintf("echo \"%s\" > /config/my-pod.yaml", strings.ReplaceAll(yamlContent.YAML, "\"", "\\\""))}
+		initContainer := workloads.NewContainer("copy-yaml", imageSetting.Value, corev1.PullAlways, initVolumeMount, nil, initContainerCommand, nil, nil)
+		jobTemplate.Spec.Template.Spec.InitContainers = append(jobTemplate.Spec.Template.Spec.InitContainers, initContainer)
+	}
 
-	command := []string{"kubectl", "apply", "-f", "/root/.kube/my-pod.yaml"}
 	container := workloads.NewContainer(jobName, imageSetting.Value, corev1.PullAlways, volumeMount, nil, command, securityContext, nil)
```
anupama2501, line 73: Probably best if we add an error check for command also. if we get an empty string for command, the function fails.

extensions/kubectl/template.go
```diff
@@ -63,7 +64,7 @@ func CreateJobAndRunKubectlCommands(clusterID, jobname string, job *batchv1.Job,
 
 	sa := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "rancher-installer2",
+			Name: namegen.AppendRandomString("rancher-installer2"),
```
anupama2501, line 67: lets change rancher-installer2 to kubectl-command or any similar wording?

dasarinaidu (approved):
lgtm.

Priyashetty17 (approved):
lgtm

extensions/kubectl/command.go
brudnak: I was seeing problems/bugs depending on this use/case amount of times this is run locally with namegen.AppendRandomString as it requires a base string. Then when searching with that on line 90.

extensions/kubectl/template.go
brudnak: Changed 👍

extensions/kubectl/command.go
brudnak: Removed, good catch! 👍

extensions/kubectl/command.go
brudnak: Added error check for empty command slice. Thanks!

anupama2501 (approved):

extensions/kubectl/command.go
```diff
@@ -18,17 +20,25 @@ import (
...
 const volumeName = "config"
 
-// Apply is a helper function that call a helper to create a Job and executing kubectl apply by copying the yaml
-// in the pod. It returns the output from the pod logs.
-func Apply(client *rancher.Client, cluster *apisV1.Cluster, yamlContent *management.ImportClusterYamlInput, clusterID string) (string, error) {
+// Command creates a kubernetes job to execute the provided commands.
+// It initializes necessary volume and volume mounts for the job and overrides securityContext to execute the commands.
+// If yamlContent is provided, an init container is created to write this content to a file.
+// After creating the job it waits for the execution of the job and retrieves the logs of the executed commands
+func Command(client *rancher.Client, cluster *apisV1.Cluster, yamlContent *management.ImportClusterYamlInput, clusterID string, command []string) (string, error) {
```
caliskanugur, line 27: Maybe I am missing it, I can't read where cluster arg is used. Is that object necessary for this function?

caliskanugur (commented):
Just have a single question, looks good to me

extensions/kubectl/command.go
brudnak: Ah good catch, thanks! Removed the unused parameter 👍

caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

brudnak:
- v2.7 => https://github.com/rancher/shepherd/pull/79
- v2.8 => https://github.com/rancher/shepherd/pull/78

PR #41, opened 2024-01-23
[v2.7] cis benchmark extension

vivek-shilimkar:
An extension for cis benchmark chart testing automation.

### DIFF
--- a/extensions/charts/ranchercisbenchmark.go
+++ b/extensions/charts/ranchercisbenchmark.go
@@ -1,0 +1,314 @@
+package charts
+
+import (
+	"context"
+	"fmt"
+
+	catalogv1 "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/clients/rancher/catalog"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+	"github.com/rancher/shepherd/pkg/wait"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
+	// Namespace that rancher CIS Benchmark chart is installed in
+	RancherCisBenchmarkNamespace = "cis-operator-system"
+	// Name of the rancher CIS Benchmark chart
+	RancherCisBenchmarkName = "rancher-cis-benchmark"
+	// Name of rancher CIS Benchmark crd chart
+	RancherCisBenchmarkCRDName = "rancher-cis-benchmark-crd"
+	CisbenchmarkProjectName    = "cis-operator-system"
+	ClusterScanResourceType    = "cis.cattle.io.clusterscan"
+	ClusterScanReportType      = "cis.cattle.io.clusterscanreport"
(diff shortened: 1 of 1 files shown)


extensions/charts/ranchercisbenchmark.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	chartInstallAction := newCisBenchmarkChartInstallAction(cisbenchmarkChartInstallActionPayload)
+
+	catalogClient, err := client.GetClusterCatalogClient(installOptions.ClusterID)
+	if err != nil {
+		return err
+	}
+
+	// Install the Rancher CIS Benchmark chart
```
slickwarren, line 98: this comment is redundant.

extensions/charts/ranchercisbenchmark.go
```diff
@@ -0,0 +1,215 @@
...
+	err = catalogClient.InstallChart(chartInstallAction, catalog.RancherChartRepo)
+	if err != nil {
+		return err
+	}
+
+	// Wait for chart to be fully deployed
```
slickwarren, line 104: I think this comment is also redundant / not necessary

extensions/charts/ranchercisbenchmark.go
```diff
@@ -0,0 +1,215 @@
...
+		if state == string(catalogv1.StatusDeployed) {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Register cleanup function for uninstallation
```
slickwarren, line 126: also is not necessary. A better comment here would be why the cleanup function isn't set elsewhere, like before chart install. in other words, why does it need to be explicitly set here vs. at the beginning of the function? i.e., what if one of the above functions returns an error. Does that mean no part of the chart was installed, and thus no cleanup is necessary? Sometimes when testing charts I've seen the crd chart installed, but the main chart fail.

extensions/charts/ranchercisbenchmark.go
```diff
@@ -0,0 +1,215 @@
...
+	client.Session.RegisterCleanupFunc(func() error {
+		return UninstallRancherCisBenchmarkChart(client, installOptions)
+	})
+
+	return nil
+}
+
+// UninstallRancherCisBenchmarkChart is a helper function that uninstalls the rancher-CIS Benchmark chart.
+func UninstallRancherCisBenchmarkChart(client *rancher.Client, installOptions *InstallOptions) error {
```
slickwarren, line 135: is there a reason this is specific to cis chart? it seems like you could enter the chart parameters as variables, and make a generic uninstallChart function that could be used instead of specific functions for each uninstall.

extensions/charts/ranchercisbenchmark.go
```diff
@@ -0,0 +1,215 @@
...
+	catalogClient, err := client.GetClusterCatalogClient(installOptions.ClusterID)
+	if err != nil {
+		return err
+	}
+
+	// Uninstall the Rancher CIS Benchmark chart
```
slickwarren, line 141: a lot of these comments seem unnecessary imo.

slickwarren (commented):
some nits on the comments, and one question about the cleanup registration function.

caliskanugur (commented):
I've got a few questions

caliskanugur:
Do we have sanity checks for this?

PR #48, opened 2024-01-30
[2.7] Remove check for Rancher version in the kubernetesversions extension

Priyashetty17:
Issue: https://github.com/rancher/qa-tasks/issues/1122

The [current implementation](https://github.com/rancher/rancher/blob/b57600cdc9b427f30f6c964ad46c535cce8f5c86/tests/framework/extensions/clusters/kubernetesversions/all.go#L228) assumes that the endpoints v1-rke2-release/releases and v1-k3s-release/releases return ALL available versions of RKE2/K3s (both compatible and incompatible). And, to filter out the compatible versions, we have code that checks if the Rancher version falls within the specified range of minChannelVersion to maxChannelVersion. I tested it with few Rancher versions and also confirmed with Max Ross that the endpoints return only the compatible versions. So, we can now remove this check. Extension 'kubernetesversions' is being updated to reflect this change.

Forward ports for 2.8 and main will also be created for this

### DIFF
--- a/extensions/clusters/kubernetesversions/all.go
+++ b/extensions/clusters/kubernetesversions/all.go
@@ -14,8 +14,6 @@ import (
 )
 
 const (
-	rancherVersionSetting = "server-version"
-
 	rke1VersionsSetting = "k8s-versions-current"
 	rke2ReleasePath     = "v1-rke2-release/releases"
 	k3sReleasePath      = "v1-k3s-release/releases"
@@ -42,15 +40,6 @@ func ListRKE1AllVersions(client *rancher.Client) (allAvailableVersions []string,
 
 // ListRKE2AllVersions is a function that uses the management client and releases endpoint to list and return all RKE2 versions.
 func ListRKE2AllVersions(client *rancher.Client) (allAvailableVersions []string, err error) {
-	setting, err := client.Management.Setting.ByID(rancherVersionSetting)
-	if err != nil {
-		return
-	}
-	rancherVersion, err := semver.NewVersion(setting.Value)
-	if err != nil {
-		return
-	}
-
 	url := fmt.Sprintf("%s://%s/%s", "http", client.RancherConfig.Host, rke2ReleasePath)
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
@@ -71,7 +60,7 @@ func ListRKE2AllVersions(client *rancher.Client) (allAvailableVersions []string,
 
 	releases := mapResponse["data"].([]interface{})
 
-	allAvailableVersions = sortReleases(rancherVersion, releases)
+	allAvailableVersions = sortReleases(releases)
 
 	sort.Strings(allAvailableVersions)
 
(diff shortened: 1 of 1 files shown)


anupama2501:
I still notice multiple kubernetes versions listed on 2.8.2 at the endpoint `v1-k3s-release/releases`. For example, on 2.8.2 I see 3 versions for 1.25:

- v1.25.15+k3s2
- v1.25.16+k3s4
- v1.25.13+k3s1

Considering the 2.8.2 version supports the k3s version `v1.25.16+k3s4`, if we remove the rancher version k8s compatibility  check we might end up picking a version of k8s that is not supported on a rancher release. 
Please correct me if my understanding is incorrect but we would need to have the rancher server check.

anupama2501:
Spoke to @Priyashetty17 offline. Issue happens with rancher head and other patch versions with an` invalid semnatic version` error. Current strategy attempts to verify the k8s version under each rancher server version but each of these k8s have the same min and max versions. For example:

extensions/clusters/kubernetesversions/all.go
```diff
@@ -42,15 +42,6 @@ func ListRKE1AllVersions(client *rancher.Client) (allAvailableVersions []string,
 
 // ListRKE2AllVersions is a function that uses the management client and releases endpoint to list and return all RKE2 versions.
 func ListRKE2AllVersions(client *rancher.Client) (allAvailableVersions []string, err error) {
-	setting, err := client.Management.Setting.ByID(rancherVersionSetting)
```
anupama2501, line 45: We can remove the constant `rancherVersionSetting` as its not referenced anywhere other than these tests.

Priyashetty17: Thanks for catching that! Removed

extensions/clusters/kubernetesversions/all.go
```diff
@@ -236,14 +215,9 @@ func sortReleases(rancherVersion *semver.Version, releases []interface{}) (allAv
 			continue
 		}
 
-		maxVersion := release.(map[string]interface{})["maxChannelServerVersion"].(string)
 		minVersion := release.(map[string]interface{})["minChannelServerVersion"].(string)
```
anupama2501, line 218: Have a qq, should we be removing the min version also from the code? Since max and min were both used to find the appropriate k8s versions?

Priyashetty17: min version might still be necessary to generate a map of available versions, as Kubernetes versions vary based on specific minimum Rancher versions.

@caliskanugur Do you have inputs on this?

caliskanugur: Great question, it should affect it, but it might've also changed like this max version, we would need to check the newest response. 

I'd vote to update this logic one step at a time with a backlog. If any changes introduce some unexpected output, we can have a clearer look at the history of this sorting function to make minimal fixes. What do you all think?

caliskanugur: https://github.com/rancher/qa-tasks/issues/1123#issuecomment-1976989360, added this to an existing related ticket, awesome question @anupama2501

caliskanugur: Just realized that the min ver check was already removed here. So removing other rancher `minVersion`related things from this function should be safe

caliskanugur: Resolving this, we will stick with this approach as it's going to be used for the mapping

caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

PR #50, opened 2024-01-31
[2.8] Remove check for Rancher version in the kubernetesversions extension

Priyashetty17:
Forward port https://github.com/rancher/shepherd/pull/48

PR #51, opened 2024-01-31
Remove check for Rancher version in the kubernetesversions extension

Priyashetty17:
Forward port https://github.com/rancher/shepherd/pull/48

Issue: https://github.com/rancher/qa-tasks/issues/1122

### DIFF
--- a/extensions/clusters/kubernetesversions/all.go
+++ b/extensions/clusters/kubernetesversions/all.go
@@ -14,8 +14,6 @@ import (
 )
 
 const (
-	rancherVersionSetting = "server-version"
-
 	rke1VersionsSetting = "k8s-versions-current"
 	rke2ReleasePath     = "v1-rke2-release/releases"
 	k3sReleasePath      = "v1-k3s-release/releases"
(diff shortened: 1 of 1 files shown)


extensions/clusters/kubernetesversions/all.go
```diff
@@ -254,10 +228,6 @@ func sortReleases(rancherVersion *semver.Version, releases []interface{}) (allAv
 			continue
 		}
 
-		if !rancherVersion.GreaterThan(minRancherVersion) && !rancherVersion.LessThan(maxRancherVersion) {
-			continue
-		}
```
caliskanugur, line 259: Priya just a small thing about this, we still want to use `minRancherVersion` so we need to ensure the current rancher version is greater than the `minRancherVersion`, so something like this:

		if !rancherVersion.GreaterThan(minRancherVersion) {
			continue
		}

caliskanugur (commented):
I've got only one comment, otherwise looks good to me

extensions/clusters/kubernetesversions/all.go
Priyashetty17: @caliskanugur I think this may not be needed as well. Based on what I have noticed, what gets listed with the endpoint are the compatible versions for that specific rancher version. If the min version is less than the rancher version , that wouldn't be listed

caliskanugur: You're right, this is also related to @anupama2501 's min version question, so from Priya's previous confirmation with the KDM team turns out we don't need `minRancherVersion` at all either. Resolving this one.

caliskanugur:
Could we get a rebase for this and its backports?

Priyashetty17:
@caliskanugur Rebased this and the backports

caliskanugur (approved):
Looks good to me

caliskanugur:
Backports:

2.7: https://github.com/rancher/shepherd/pull/48
2.8: https://github.com/rancher/shepherd/pull/50

igomez06 (approved):
LGTM

PR #53, opened 2024-01-31
[v2.8] Fix etcdctl version check for static and dynamic tests

markusewalker:
### PR Description
When running the RKE1 custom cluster provisioning tests, there is a specific check for the `etcdctl` version. When running the DynamicInput tests, this passes. For the static TDD tests, they fail.

Troubleshooting, it is because the etcd node has the private IP address in the external AND internal IP address for only the static tests; the dynamic tests do not share this problem.

### Solution
To fix this/polish up the check further, removing the additional if condition for checking the externalIP; so long as the `etcdRole` is true, run the docker exec command. This will occur only on the first etcd node, then it will break the check.


Josh-Diamond (approved):

slickwarren (approved):

caliskanugur (approved):
Looks good to me

Josh-Diamond (approved):

slickwarren (approved):

igomez06 (approved):
LGTM

PR #62, opened 2024-02-12
[2.7] Adding extensions for rbac checks

anupama2501:
Back port: https://github.com/rancher/shepherd/pull/16

### DIFF
--- a/extensions/kubeapi/rbac/list.go
+++ b/extensions/kubeapi/rbac/list.go
@@ -32,7 +32,6 @@ func ListRoleBindings(client *rancher.Client, clusterName, namespace string, lis
 
 		rbList.Items = append(rbList.Items, *rb)
 	}
-
 	return rbList, nil
 }
 
@@ -55,7 +54,6 @@ func ListClusterRoleBindings(client *rancher.Client, clusterName string, listOpt
 		if err != nil {
 			return nil, err
 		}
-
 		crbList.Items = append(crbList.Items, *crb)
 	}
 
@@ -81,7 +79,6 @@ func ListGlobalRoleBindings(client *rancher.Client, listOpt metav1.ListOptions) 
 		if err != nil {
 			return nil, err
 		}
-
 		grbList.Items = append(grbList.Items, *grb)
 	}
 
@@ -107,7 +104,6 @@ func ListClusterRoleTemplateBindings(client *rancher.Client, listOpt metav1.List
 		if err != nil {
 			return nil, err
 		}
-
 		crtbList.Items = append(crtbList.Items, *crtb)
 	}
 
@@ -133,7 +129,6 @@ func ListGlobalRoles(client *rancher.Client, listOpt metav1.ListOptions) (*v3.Gl
 		if err != nil {
 			return nil, err
 		}
-
 		grList.Items = append(grList.Items, *gr)
 	}
 
@@ -159,7 +154,6 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
 		if err != nil {
 			return nil, err
 		}
-
 		rtList.Items = append(rtList.Items, *rt)
 	}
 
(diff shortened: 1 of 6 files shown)


igomez06 (approved):
LGTM

anupama2501:
@caliskanugur Updated this PR as well.

Priyashetty17 (approved):
lgtm

extensions/kubeapi/rbac/rbac.go
```diff
@@ -66,3 +72,32 @@ var RoleTemplateGroupVersionResource = schema.GroupVersionResource{
 	Version:  Version,
 	Resource: "roletemplates",
 }
+
+// ProjectRoleTemplateBindingGroupVersionResource is the required Group Version Resource for accessing projectroletemplatebindings in a cluster, using the dynamic client.
+var ProjectRoleTemplateBindingGroupVersionResource = schema.GroupVersionResource{
```
caliskanugur, line 77: Is this diff expected? This is something new for 2.7 right? It seems like 2.8 and 2.9 already have it

caliskanugur (commented):
Looks good, I just have one question

extensions/kubeapi/rbac/rbac.go
Priyashetty17: @anupama2501 I realized this only after reading Ugur's question. Since Public API was introduced in 2.8, we cannot have the following PRTB related methods in 2.7, correct?
- ListProjectRoleTemplateBindings
- GetProjectRoleTemplateBindingsByName

extensions/kubeapi/rbac/rbac.go
```diff
@@ -66,3 +72,32 @@ var RoleTemplateGroupVersionResource = schema.GroupVersionResource{
...
+	Group:    GroupName,
+	Version:  Version,
+	Resource: "projectroletemplatebindings",
+}
+
+
+// GetProjectRoleTemplateBindingsByName is a helper function that uses the dynamic client to get a projectroletemplatebinding from local cluster.
+func GetProjectRoleTemplateBindingsByName(client *rancher.Client, getOpts metav1.GetOptions, name string) (*v3.ProjectRoleTemplateBinding, error) {
```
Priyashetty17, line 85: Since Public API was introduced in 2.8, can you remove this for 2.7?

extensions/kubeapi/rbac/list.go
```diff
@@ -159,9 +154,34 @@ func ListRoleTemplates(client *rancher.Client, listOpt metav1.ListOptions) (*v3.
...
 		}
-
 		rtList.Items = append(rtList.Items, *rt)
 	}
 
 	return rtList, nil
 }
+
+// ListProjectRoleTemplateBindings is a helper function that uses the dynamic client to list projectroletemplatebindings from local cluster.
+func ListProjectRoleTemplateBindings(client *rancher.Client, listOpt metav1.ListOptions) (*v3.ProjectRoleTemplateBindingList, error) {
```
Priyashetty17, line 164: Since Public API was introduced in 2.8, can you remove this for 2.7?

extensions/kubeapi/rbac/rbac.go
igomez06: Do you still want this Priya? Should we wait on this PR until that is removed?

Priyashetty17: Israel, I think we should wait until this is removed for 2.7

extensions/kubeapi/rbac/rbac.go
anupama2501: ah its going to be a bit tricky. So, for the tests that are using these functions, we would need to update them to use v3 or v1 API specifically for 2.7 and keep 2.8 and 2.9 use public API.

anupama2501: Removed public API references here. thank you for calling these out!

Priyashetty17 (approved):
lgtm

caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

PR #64, opened 2024-02-14
adds support for deleting init node

Josh-Diamond:
extensions to support: https://github.com/rancher/rancher/pull/44155

### DIFF
--- a/extensions/nodes/nodes.go
+++ b/extensions/nodes/nodes.go
@@ -1,0 +1,92 @@
+package nodes
+
+import (
+	"errors"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
+	"github.com/rancher/shepherd/extensions/kubeapi/secrets"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	initNodeLabelKey      = "rke.cattle.io/init-node"
+	local                 = "local"
+	machinePlanSecretType = "rke.cattle.io/machine-plan"
+	machineNameSteveLabel = "rke.cattle.io/machine-name"
+)
+
+// GetInitMachine accepts a client and clusterID and returns the "init node" machine
+// object for rke2/k3s clusters
+func GetInitMachine(client *rancher.Client, clusterID string) (*v1.SteveAPIObject, error) {
+	logrus.Info("Retrieving list of secrets...")
+	secretList, err := secrets.ListSecrets(client, local, fleetNamespace, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	machinePlans := []corev1.Secret{}
+
+	logrus.Info("Identifying machine-plan secrets...")
+	for _, secret := range secretList.Items {
+		if secret.Type == machinePlanSecretType {
+			logrus.Infof("Found machine-plan secret: %s", secret.Name)
+			machinePlans = append(machinePlans, secret)
+		}
+	}
+
+	initNodeMachineName := ""
+
+	logrus.Info("Identifying init node machine plan...")
+	for _, machinePlan := range machinePlans {
+		v, ok := machinePlan.ObjectMeta.Labels[initNodeLabelKey]
+		if ok && v == "true" {
+			logrus.Infof("Found init node machine plan: %s", machinePlan.Name)
+			initNodeMachineName = machinePlan.Labels[machineNameSteveLabel]
+		}
+	}
+
+	logrus.Info("Retrieving machine...")
+	initMachine, err := client.Steve.SteveType(machineSteveResourceType).ByID(fleetNamespace + "/" + initNodeMachineName)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Successfully retrieved machine: %s", initNodeMachineName)
+
+	return initMachine, nil
+}
+
+// DeleteMachineRKE2K3S accepts a client and v1.SteveAPIObject and deletes the machine
+func DeleteMachineRKE2K3S(client *rancher.Client, machine *v1.SteveAPIObject) error {
+	logrus.Info("Deleting machine...")
+	err := client.Steve.SteveType(machineSteveResourceType).Delete(machine)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// VerifyDeletedMachineRKE2K3S accepts a client and v1.SteveAPIObject and verifies that
+// the machine was successfully removed from the cluster
+func VerifyDeletedMachineRKE2K3S(client *rancher.Client, deletedMachine *v1.SteveAPIObject) error {
+	logrus.Info("Re-fetching machines...")
+
+	machines, err := client.Steve.SteveType(machineSteveResourceType).ListAll(nil)
+	if err != nil {
+		return err
+	}
+
+	logrus.Info("Verifying machine was successfully removed...")
+	for _, machine := range machines.Data {
+		if machine.Resource.ID == deletedMachine.Resource.ID {
+			return errors.New("Machine was not successfully removed from cluster")
+		}
+	}
+
+	logrus.Info("Machine was successfully removed from cluster!")
+	return nil
+}


extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+// DeleteNodeRKE2K3S() accepts a client and node object and deletes the node from the cluster
+func DeleteNodeRKE2K3S(client *rancher.Client, node *v3.Node) error {
+	logrus.Info("Targeting node for deletion...")
+	machine, err := client.Steve.SteveType(machineSteveResourceType).ByID(fleetNamespace + "/" + node.Annotations[machineSteveAnnotation])
+	if err != nil {
+		return  err
+	}
+
+	logrus.Info("Deleting node...")
+	err = client.Steve.SteveType(machineSteveResourceType).Delete(machine)
```
igomez06, line 88: So it's delete a cluster.x-k8s.io.machine not the node?

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return  err
+	}
+
+	return nil
+}
+
+// VerifyDeletedNodeRKE2K3S() accepts a client and node object and verifies that the node was successfully removed from the cluster
+func VerifyDeletedNodeRKE2K3S(client *rancher.Client, node *v3.Node) error {
```
igomez06, line 97: Do we really need the verify here? Shouldn't this just be in the test?

igomez06 (commented):
A few questions.

extensions/nodes/nodes.go
Josh-Diamond: I felt this logic was re-usable and could be useful in future tests, so in an effort to keep things DRY moving forward, I broke it out into an extension.  Upon reviewing this, I'm noticing the logs are specific to my `init node` task, so I will need to update them to be more generic, unless its requested to remove this from the extensions altogether.

extensions/nodes/nodes.go
Josh-Diamond: Yes.  The function accepts a v3.Node object, but its essentially targeting and deleting a machine.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+const (
+	initNodeLabelKey         = "rke.cattle.io/init-node"
+	initNodeLabelValue       = "true"
+	local                    = "local"
+	machinePlanSecretType    = "rke.cattle.io/machine-plan"
+	machineNameSteveLabel    = "rke.cattle.io/machine-name"
+	node                     = "node"
+)
+
+// GetInitNode() accepts a client and clusterID and returns the init node object
```
markusewalker, line 22: nit - can the comment just say `GetInitNode` and not `GetInitNode()`.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	logrus.Info("Retrieving init node...")
+	initNode, err := client.Management.Node.ByID(initNodeID)
+	if err != nil {
+		return nil, err
+	}
+
+	return initNode, nil
+}
+
+// DeleteNodeRKE2K3S() accepts a client and node object and deletes the node from the cluster
```
markusewalker, line 79: Same nit here.

extensions/nodes/nodes.go
markusewalker: This function as well is more generic as opposed to searching for the init node specifically.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	logrus.Info("Deleting node...")
+	err = client.Steve.SteveType(machineSteveResourceType).Delete(machine)
+	if err != nil {
+		return  err
+	}
+
+	return nil
+}
+
+// VerifyDeletedNodeRKE2K3S() accepts a client and node object and verifies that the node was successfully removed from the cluster
```
markusewalker, line 96: Same nit here.

markusewalker (commented):
Overall, looks good. Left few comments.

extensions/nodes/nodes.go
Josh-Diamond: Done!

extensions/nodes/nodes.go
Josh-Diamond: Done!

extensions/nodes/nodes.go
Josh-Diamond: Done!

slickwarren (approved):
lgtm

markusewalker (commented):
Overall, LGTM. Can you make sure that you run this PR against the linter the team discussed and adjust accordingly to its feedback?

Also, once this PR is approved, please open the backport PRs.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	
+	logrus.Info("Identifying init node machine plan...")
+	for _, machinePlan := range machinePlans {
+		for key, val := range machinePlan.ObjectMeta.Labels {
+			if key == initNodeLabelKey && val == initNodeLabelValue {
+				logrus.Infof("Found init node machine plan: %s", machinePlan.Name)
+				initNodeMachineName = machinePlan.Labels[machineNameSteveLabel]
+			}
+		}
+	}
```
caliskanugur, line 50: A question here, does it need to create nested for loops? It can check map keys, something like this;

```suggestion
	for _, machinePlan := range machinePlans {
	        v, ok := machinePlan.ObjectMeta.Labels[initNodeLabelKey]
	        if ok && v ==  initNodeLabelValue {
	             		logrus.Infof("Found init node machine plan: %s", machinePlan.Name)
				initNodeMachineName = machinePlan.Labels[machineNameSteveLabel]
		}
	}
```

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	initNodeID := ""
+
+	logrus.Info("Identifying init node...")
+	for _, node := range nodes.Data {
+		for key, val := range node.Annotations {
+			if key == machineSteveAnnotation && val == initNodeMachineName {
+				logrus.Infof("Found init node: %s", node.ID)
+				initNodeID = node.ID
+			}
+		}
```
caliskanugur, line 67: The same applies here too, could you check if the map contains a key rather than creating a loop for the map?

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+
+	nodes, err := client.Management.Node.ListAll(&types.ListOpts{})
+	if err != nil {
+		return err
+	}
+
+	logrus.Info("Verifying node was successfully removed...")
+	for _, machine := range nodes.Data {
+		if machine.ID == node.ID {
+			logrus.Fatal("Node was not successfully removed from cluster")
```
igomez06, line 108: Why is this a fatal. We should never have fatals ever. It completely exits the test, all tests cancel. We don't want that.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return nil
+}
+
+// VerifyDeletedNodeRKE2K3S() accepts a client and node object and verifies that the node was successfully removed from the cluster
+func VerifyDeletedNodeRKE2K3S(client *rancher.Client, node *v3.Node) error {
+	logrus.Info("Re-fetching nodes...")
+
+	nodes, err := client.Management.Node.ListAll(&types.ListOpts{})
```
igomez06, line 100: My only other nit, is new code should use v1 calls and stop using v3/norman. Unless you can convince me otherwise.

extensions/nodes/nodes.go
igomez06: Why does it accept a v3.node instead of a v1 node?

caliskanugur (commented):
I've got a few questions

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/extensions/kubeapi/secrets"
+	corev1 "k8s.io/api/core/v1"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	initNodeLabelKey         = "rke.cattle.io/init-node"
+	initNodeLabelValue       = "true"
```
caliskanugur, line 15: I think this is not necessary it can be a parse `b, err := strconv.ParseBool("true")` something like this once it gets the label? This would happen in the code snippet I commented below

extensions/nodes/nodes.go
Josh-Diamond: I've updated this to use v1

extensions/nodes/nodes.go
Josh-Diamond: I removed the fatal

extensions/nodes/nodes.go
Josh-Diamond: great suggestion! thank you Ugur! i've implemented this

markusewalker (commented):
Linter caught two issues. Will approve once they have been addressed.

markusewalker (approved):
LGTM

susesgartner (approved):
LGTM

slickwarren (approved):

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,92 @@
...
+			machinePlans = append(machinePlans, secret)
+		}
+	}
+
+	initNodeMachineName := ""
+
+	logrus.Info("Identifying init node machine plan...")
+	for _, machinePlan := range machinePlans {
+		v, ok := machinePlan.ObjectMeta.Labels[initNodeLabelKey]
+		if ok && v == "true" {
```
caliskanugur, line 45: Josh, could you please check this: https://github.com/rancher/shepherd/pull/64#discussion_r1525497184. You can find more info on offline PR guidelines docs (15th)

Josh-Diamond: Hey Ugur,

I reviewed the comment, and when implementing it, I was getting a type error where bool cannot be type string, so I hardcoded this to "true" as it should never change. I felt this resolved the initial concerns with using the const and is a simple and clean resolution.

Do you have any concerns with this approach? 🙂

PR #67, opened 2024-02-16
[v2.8] Fixed issue with post install steps from install Rancher

igomez06:
Moved steveClient instantiation into the poll, to make sure the management.cattle.io.setting schema is properly found. This has been causing intermittent issues in recurring runs.


floatingman (approved):

caliskanugur (approved):
Looks good to me

PR #74, opened 2024-02-21
make out-of-tree helpers public for additional testing during upgrades

slickwarren:
note; sometimes, the kubectl command is still flaky. This was pulled from this [PR](https://github.com/rancher/shepherd/pull/17) since it hasn't merged yet, which is an improvement from what's currently in but not 100% 

example of the bug from kubectl, hit intermittently
```
an error on the server ("unable to create impersonator account: error setting up impersonation for user user-kcbbw: impersonation: error ensuring secret for service account cattle-impersonation-user-kcbbw: error ensuring secret for service account [cattle-impersonation-system:cattle-impersonation-user-kcbbw]: timed out waiting for the condition") has prevented the request from succeeding
```

the other problem I'll see is this, followed by a timeout
```
child pods are preserved by default when jobs are deleted; set propagationPolicy=Background to remove them or set propagationPolicy=Orphan to suppress this warning
```

### DIFF
--- a/extensions/charts/awsoutoftree.go
+++ b/extensions/charts/awsoutoftree.go
@@ -23,7 +23,7 @@ const (
 )
 
 // InstallAWSOutOfTreeChart installs the CSI chart for aws cloud provider in a given cluster.
-func InstallAWSOutOfTreeChart(client *rancher.Client, installOptions *InstallOptions, repoName, clusterID string) error {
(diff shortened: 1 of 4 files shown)


markusewalker (approved):
LGTM.

Josh-Diamond (approved):

igomez06 (dismissed):
LGTM, let's coordinate the merge so we can have it working on rancher/rancher

PR #79, opened 2024-02-26
[v2.7] Refactor Apply function to Command

brudnak:
Updated kubectl apply function to command.
Also removed kubectl explain function, now that
command can handle any kubectl situation.

Added in option buffer size parsing to
kubeconfig pod logs. This is needed to gather
large logs with kubectl logs cmd.

Fixed a bug that was formally in kubectl
apply.


igomez06 (approved):
LGTM

PR #92, opened 2024-03-07
[v2.7] v2.7.11 version bumps

igomez06:
There will be version bumps that are coming from the v2.7.11 release that we need to address in shepherd.


floatingman (approved):

caliskanugur (approved):
Looks good to me

PR #97, opened 2024-03-08
Add extension to verify role bindings for user 

joesims22:
This PR adds the extension VerifyRoleBindingsForUser which validates the corresponding role bindings have been created for a user.

### DIFF
--- a/extensions/rbac/verify.go
+++ b/extensions/rbac/verify.go
@@ -12,6 +12,7 @@ import (
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/clusters"
+	"github.com/rancher/shepherd/extensions/kubeapi/rbac"
 	"github.com/rancher/shepherd/extensions/namespaces"
 	"github.com/rancher/shepherd/extensions/projects"
 	"github.com/rancher/shepherd/extensions/users"
@@ -20,6 +21,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	coreV1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -34,6 +36,8 @@ const (
 	standardUser                  = "user"
 	activeStatus                  = "active"
 	forbiddenError                = "403 Forbidden"
+	localCluster                  = "local"
+	userKind                      = "User"
 )
 
 var rgx = regexp.MustCompile(`\[(.*?)\]`)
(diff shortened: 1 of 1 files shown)


extensions/rbac/verify.go
```diff
@@ -46,6 +49,27 @@ func VerifyGlobalRoleBindingsForUser(t *testing.T, user *management.User, adminC
 	assert.Equal(t, 1, len(grbs.Data))
 }
 
+// VerifyRoleBindingsForUser validates that the corresponding role bindings are created for the user
+func VerifyRoleBindingsForUser(t *testing.T, user *management.User, standardClient *rancher.Client, clusterID, role string) {
+	rblist, err := rbac.ListRoleBindings(standardClient, localcluster, clusterID, metav1.ListOptions{})
```
anupama2501, line 54: shouldn't the client be admin client? standard user client won't be able to list role bindings from the local cluster.

extensions/rbac/verify.go
```diff
@@ -46,6 +49,27 @@ func VerifyGlobalRoleBindingsForUser(t *testing.T, user *management.User, adminC
...
+	require.NoError(t, err)
+	userID := user.Resource.ID
+	userRoleBindings := []string{}
+
+	for _, rb := range rblist.Items {
+		if rb.Subjects[0].Kind == "User" && rb.Subjects[0].Name == userID {
```
anupama2501, line 60: Can we add "User" as a constant?

anupama2501 (commented):
looks good, have a few comments. Do we have a jenkins log for this ?

extensions/rbac/verify.go
joesims22: Yes that's correct, I'll change it to admin.

extensions/rbac/verify.go
joesims22: Will add User as a constant 👍

extensions/rbac/verify.go
```diff
@@ -34,6 +36,8 @@ const (
 	standardUser                  = "user"
 	activeStatus                  = "active"
 	forbiddenError                = "403 Forbidden"
+	localcluster                  = "local"
```
Priyashetty17, line 39: nit., Can you change the variable name to `localCluster`?

Priyashetty17 (commented):
lgtm, just a nit on the variable name

extensions/rbac/verify.go
joesims22: changed constant name 👍

anupama2501 (approved):
lgtm

extensions/rbac/verify.go
```diff
@@ -46,6 +50,27 @@ func VerifyGlobalRoleBindingsForUser(t *testing.T, user *management.User, adminC
 	assert.Equal(t, 1, len(grbs.Data))
 }
 
+// VerifyRoleBindingsForUser validates that the corresponding role bindings are created for the user
+func VerifyRoleBindingsForUser(t *testing.T, user *management.User, adminClient *rancher.Client, clusterID, role string) {
+	rblist, err := rbac.ListRoleBindings(adminClient, localcluster, clusterID, metav1.ListOptions{})
```
Priyashetty17, line 55: Can you update the constant here?

joesims22: updated 👍

Priyashetty17 (approved):
lgtm

igomez06 (commented):
My question for this and maybe @anupama2501 can help answer it. Why do we need this to be a verification in the framework. Can't it just go in the test itself?

igomez06:
> @igomez06 this was part of the rbac validations we moved from validation suite to extension suite as part of this task [rancher/qa-tasks#892](https://github.com/rancher/qa-tasks/issues/892)

This should only be the case if the verification is being used outside of the package, if not we're unnecessarily bloating the framework when this func should just be in the test. So you're saying VerifyRoleBindingsForUser is being used outside of tests/v2/validation/rbac?

anupama2501:
sorry @igomez06 for the delayed response. Yes, this is to make use of the role bindings verification outside of `VerifyRoleBindingsForUser` as well. We have a few tests in rbac additional for which we can make use of this function.

However, considering this involves actions rather than verification, it is appropriate for actions. Since we already have a PR for this, we can merge it now and later move it to actions once that is established. Let me know if that sounds good.

igomez06:
> sorry @igomez06 for the delayed response. Yes, this is to make use of the role bindings verification outside of `VerifyRoleBindingsForUser` as well. We have a few tests in rbac additional for which we can make use of this function.
> 
> However, considering this involves actions rather than verification, it is appropriate for actions. Since we already have a PR for this, we can merge it now and later move it to actions once that is established. Let me know if that sounds good.

Yes this would be moved any way, but my question remains. Is this function going to be used outside of tests/v2/validation/rbac? If not, then I would argue that this should just go into tests/v2/validation/rbac/rbac.go

anupama2501:
Yes, once the RBAC tests are relocated to the integration suite, they will be referenced outside of the tests/v2/validation/ directory. In that scenario, this should be maintained in actions.

igomez06 (approved):
LGTM

anupama2501:
Yes @igomez06 will wait on actions to be implemented and once thats done this can be ported over to actions. We can then close this PR.

joesims22:
This code has been ported over to actions [#47447](https://github.com/rancher/rancher/pull/47447)

Closing this since code will reside in rancher/rancher actions.

PR #103, opened 2024-03-14
Enable automated install of vsphere cpi/csi

slickwarren:
https://github.com/rancher/rancher/pull/44787

enables the following option in the config:
```
provisioningInput:
  cloudProvider: "rancher-vsphere"
```

which will allow the user to use vsphere with either rke1 or rke2. 

Other Improvements:
* added an option to populate default values for a given chart
* formatting updates
* added `DatastoreUrl` for vsphere rke2 and rke1

### DIFF
--- a/extensions/charts/vsphereoutoftree.go
+++ b/extensions/charts/vsphereoutoftree.go
@@ -1,0 +1,178 @@
+package charts
+
+import (
+	"github.com/rancher/shepherd/extensions/projects"
+	"github.com/rancher/shepherd/extensions/rke1/nodetemplates"
+	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/clients/rancher/catalog"
+	r1vsphere "github.com/rancher/shepherd/extensions/rke1/nodetemplates/vsphere"
+)
+
+const (
+	systemProject       = "System"
+	vsphereCPIchartName = "rancher-vsphere-cpi"
+	vsphereCSIchartName = "rancher-vsphere-csi"
+
+	vcenter      = "vCenter"
+	storageclass = "storageClass"
+
+	datacenters  = "datacenters"
+	host         = "host"
+	password     = "password"
+	username     = "username"
+	port         = "port"
+	clusterid    = "clusterId"
+	datastoreurl = "datastoreURL"
+)
+
+// InstallVsphereOutOfTreeCharts installs the CPI and CSI chart for aws cloud provider in a given cluster.
+func InstallVsphereOutOfTreeCharts(client *rancher.Client, vsphereTemplate *nodetemplates.NodeTemplate, repoName, clusterID string) error {
+
+	serverSetting, err := client.Management.Setting.ByID(serverURLSettingID)
+	if err != nil {
+		return err
+	}
+
+	registrySetting, err := client.Management.Setting.ByID(defaultRegistrySettingID)
+	if err != nil {
+		return err
+	}
+
+	project, err := projects.GetProjectByName(client, clusterID, systemProject)
+	if err != nil {
+		return err
+	}
+
+	catalogClient, err := client.GetClusterCatalogClient(clusterID)
+	if err != nil {
+		return err
+	}
+
+	latestCPIVersion, err := catalogClient.GetLatestChartVersion(vsphereCPIchartName, catalog.RancherChartRepo)
+	if err != nil {
+		return err
+	}
+
+	installCPIOptions := &InstallOptions{
+		ClusterID: clusterID,
+		Version:   latestCPIVersion,
+		ProjectID: project.ID,
+	}
+
+	chartInstallActionPayload := &payloadOpts{
+		InstallOptions:  *installCPIOptions,
+		Name:            vsphereCPIchartName,
+		Namespace:       kubeSystemNamespace,
+		Host:            serverSetting.Value,
+		DefaultRegistry: registrySetting.Value,
+	}
+
+	chartInstallAction, err := vsphereCPIChartInstallAction(catalogClient,
+		chartInstallActionPayload, vsphereTemplate, installCPIOptions, repoName, kubeSystemNamespace)
+	if err != nil {
+		return err
+	}
+
+	err = catalogClient.InstallChart(chartInstallAction, repoName)
+	if err != nil {
+		return err
+	}
+
+	err = VerifyChartInstall(catalogClient, kubeSystemNamespace, vsphereCPIchartName)
+	if err != nil {
+		return err
+	}
+
+	latestCSIVersion, err := catalogClient.GetLatestChartVersion(vsphereCSIchartName, catalog.RancherChartRepo)
+	if err != nil {
+		return err
+	}
+
+	installCSIOptions := &InstallOptions{
+		ClusterID: clusterID,
+		Version:   latestCSIVersion,
+		ProjectID: project.ID,
+	}
+
+	chartInstallActionPayload = &payloadOpts{
+		InstallOptions:  *installCSIOptions,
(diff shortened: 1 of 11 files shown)


extensions/provisioning/creates.go
```diff
@@ -120,6 +125,35 @@ func CreateProvisioningCluster(client *rancher.Client, provider Provider, cluste
...
+			"username":    cloudCredential.VmwareVsphereConfig.Username,
+		}
+		clustersConfig.AddOnConfig = &provisioninginput.AddOnConfig{
+			ChartValues: &rkev1.GenericMap{
+				Data: map[string]interface{}{
+					"rancher-vsphere-cpi": map[string]interface{}{
+						"vCenter": vcenterCredentials,
+					},
+					"rancher-vsphere-csi": map[string]interface{}{
+						// "csiWindowsSupport": map[string]interface{}{
```
Josh-Diamond, line 144: Was this intended to be uncommented or cleaned up?

Josh-Diamond (commented):
left one comment -- overall this looks great

markusewalker (approved):
Outside of Josh's comment, LGTM.

extensions/provisioning/creates.go
slickwarren: let me remove that for now - there is a separate issue with windows right now blocking these charts from properly installing. Once that's in, I'll have to open a new PR with windows support.

extensions/cloudcredentials/vsphere/create.go
```diff
@@ -27,3 +27,11 @@ func CreateVsphereCloudCredentials(rancherClient *rancher.Client) (*cloudcredent
 	}
 	return resp, nil
 }
+
+func GetVspherePassword() string {
```
igomez06, line 31: We need a go docs comment here

extensions/rke1/nodetemplates/vsphere/create.go
```diff
@@ -24,15 +25,34 @@ func CreateVSphereNodeTemplate(rancherClient *rancher.Client) (*nodetemplates.No
...
 	err = rancherClient.Management.APIBaseClient.Ops.DoCreate(management.NodeTemplateType, *nodeTemplateFinal, resp)
+
 	if err != nil {
 		return nil, err
 	}
+
 	return resp, nil
 }
+
+func GetVsphereDatastoreUrl() string {
```
igomez06, line 44: missing comment here.

extensions/rke1/nodetemplates/vsphere/create.go
```diff
@@ -24,15 +25,34 @@ func CreateVSphereNodeTemplate(rancherClient *rancher.Client) (*nodetemplates.No
...
+	var vmwarevsphereNodeTemplateConfig nodetemplates.VmwareVsphereNodeTemplateConfig
+
+	config.LoadConfig(nodetemplates.VmwareVsphereNodeTemplateConfigurationFileKey, &vmwarevsphereNodeTemplateConfig)
+
+	return vmwarevsphereNodeTemplateConfig.DatastoreUrl
+}
+
+func GetVspherePassword() string {
```
igomez06, line 52: same here

igomez06: read my comment on the other get vsphere password func.

extensions/cloudcredentials/vsphere/create.go
slickwarren: the password is not actually used in the config, rather it is passed to a secret and the secret is referenced on the object. So when you 'create' the cloud credential, the password isn't viewable or retrievable directly from the object from rancher. With this new function, I can retrieve the password from the Go object instead.

igomez06: So then why is it reading from the config. that bit is confusing

slickwarren: updated the comment, yes this is reading from the config not from a go object. Because we don't return the go object in the CreateVsphereCloudCredentials function, there's no way to get it from there. So I have to read from the config again.

igomez06: Awwww! I get it sorry, this is the cloud cred not the node template proceed lol

igomez06 (commented):
One last style nit, since we just had a meeting about it

extensions/rke1/nodetemplates/vsphere/create.go
slickwarren: updated!

igomez06 (approved):
LGTM

extensions/charts/vsphereoutoftree.go
```diff
@@ -0,0 +1,178 @@
...
+	chartInstallAction, err := vsphereCPIChartInstallAction(catalogClient,
+		chartInstallActionPayload, vsphereTemplate, installCPIOptions, repoName, kubeSystemNamespace)
+	if err != nil {
+		return err
+	}
+
+	err = catalogClient.InstallChart(chartInstallAction, repoName)
+	if err != nil {
+		return err
+	}
```
caliskanugur, line 81: This one doesn't need any cleanup registration correct? 

Cause basically if these charts are uninstalled the cluster won't be functional and there's no case where it's out-of-tree and without these charts right?

extensions/charts/vsphereoutoftree.go
```diff
@@ -0,0 +1,178 @@
...
+	datacenters  = "datacenters"
+	host         = "host"
+	password     = "password"
+	username     = "username"
+	port         = "port"
+	clusterid    = "clusterId"
+	datastoreurl = "datastoreURL"
+)
+
+// InstallVsphereOutOfTreeCharts installs the CPI and CSI chart for aws cloud provider in a given cluster.
```
caliskanugur, line 30: My other question here; installing these charts separated and one by one doesn't make sense right? Both need to be installed?

extensions/provisioning/creates.go
```diff
@@ -120,6 +125,32 @@ func CreateProvisioningCluster(client *rancher.Client, provider Provider, cluste
 
 	machinePools := machinepools.
 		CreateAllMachinePools(machineConfigs, pools, machinePoolResponses, provider.Roles, hostnameTruncation)
+
+	if clustersConfig.CloudProvider == provisioninginput.VsphereCloudProviderName.String() {
+
+		vcenterCredentials := map[string]interface{}{
+			"datacenters": machinePoolConfigs[0].Object["datacenter"],
```
caliskanugur, line 132: This one is always the same right, so I guess my question is out of this PR scope, does it have to index it, or does it have to be in the array?

caliskanugur (commented):
Looks good to me, I've got a few questions

extensions/charts/vsphereoutoftree.go
```diff
@@ -0,0 +1,178 @@
...
+
+	catalogClient, err := client.GetClusterCatalogClient(clusterID)
+	if err != nil {
+		return err
+	}
+
+	latestCPIVersion, err := catalogClient.GetLatestChartVersion(vsphereCPIchartName, catalog.RancherChartRepo)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
slickwarren: correct.

extensions/charts/vsphereoutoftree.go
slickwarren: correct

extensions/provisioning/creates.go
slickwarren: it just has to be in the array

extensions/charts/vsphereoutoftree.go
slickwarren: correct, both need to be installed.

caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

PR #107, opened 2024-03-18
Constants update

susesgartner:
Update to remove several global constants and generally rework how we store/utilize certain constants that appear across many files in both rancher/rancher and rancher/shepherd. 

Motivation: The overall goal here is to streamline how we declare and use constants, this should result in fewer global vars, less importing/declaring of generic constants in specific packages, and generally more convenient code writing since generic constants can be found and imported from one place.


susesgartner:
Note to reviewers:
For this PR I have validated that the changes should be fine by doing a go build ./... in my local shepherd branch.

When reviewing please keep the following in mind:
1. There should be NO logic changes 
2. Any constant that can be exported should NOT be deleted in this PR since it might be exported to rancher/rancher. I will make 2 follow up PRs: 1 for removing constants from repos that depend on shepherd and then another to remove the constants from shepherd.
3. If there are constants that you know of that I have missed or you want added to these files feel free to send them my way and i'll add them.

extensions/kubeapi/workloads/daemonsets/create.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/rancher/shepherd/clients/rancher"
+	deflabels "github.com/rancher/shepherd/extensions/defaults/labels"
```
slickwarren, line 8: I would vote to not shorten the name, and just call it defaultLabels or something so that its more clear

extensions/pipeline/setup.go
```diff
@@ -110,7 +111,7 @@ func UpdateEULA(adminClient *rancher.Client) error {
 
 	urlSetting.Value = fmt.Sprintf("https://%s", adminClient.RancherConfig.Host)
 
-	_, err = steveClient.SteveType("management.cattle.io.setting").Update(urlSettingResp, urlSetting)
+	_, err = steveClient.SteveType(stevetypes.Setting).Update(urlSettingResp, urlSetting)
```
slickwarren, line 114: For posterity, are you sure there's no SteveSetting equivalent? If not, this is fine. Otherwise I would argue this needs to be stevetypes.ManagementSetting

extensions/sshkeys/downloadsshkeys.go
```diff
@@ -18,7 +19,7 @@ const (
 // the ssh key for a particular node.
 func DownloadSSHKeys(client *rancher.Client, machinePoolNodeName string) ([]byte, error) {
 	machinePoolNodeNameName := fmt.Sprintf("fleet-default/%s", machinePoolNodeName)
-	machine, err := client.Steve.SteveType(ClusterMachineConstraintResourceSteveType).ByID(machinePoolNodeNameName)
+	machine, err := client.Steve.SteveType(stevetypes.Machine).ByID(machinePoolNodeNameName)
```
slickwarren, line 22: just double checking, this was the same thing as `ClusterMachineConstraintResourceSteveType`?

slickwarren (dismissed):
some small nits but this is actually awesome!

markusewalker (approved):
Overall, LGTM just a reminder for myself that there are some annotations in parallel PRs that will require updating here.

susesgartner:
I will make a final pass through shepherd and update any new consts that get merged in before this PR.

extensions/pipeline/setup.go
susesgartner: Not that I know of but I agree that ManagementSetting is much more clear. I will change it for clarity.

extensions/sshkeys/downloadsshkeys.go
susesgartner: I just double checked, yes they are the same.

igomez06:
@susesgartner can you also add the globalconstant linter as well? And see if it passes that?

susesgartner:
The globalconstant linter was what sparked my interest in creating this PR. While I have removed most of the global constants there is 1 remaining that I am currently looking into.

igomez06:
> The globalconstant linter was what sparked my interest in creating this PR. While I have removed most of the global constants there is 1 remaining that I am currently looking into.

I know, what I'm saying is can you add this to our GHA linter? Since you're already doing the work it would be a great time to add it our configuration and have it pass the official verify

susesgartner:
Sure!

slickwarren (dismissed):
lgtm

markusewalker (dismissed):

extensions/defaults/schema/groupversionresources/groupversionresources.go
```diff
@@ -0,0 +1,276 @@
...
+
+func CronJob() schema.GroupVersionResource {
+	cronJob := schema.GroupVersionResource{
+		Group:    "batch",
+		Version:  "v1beta1",
+		Resource: "cronjobs",
+	}
+
+	return cronJob
+}
```
caliskanugur, line 276: Do these have to be separated? I think this file would get longer whenever a new GVR is used in the framework.

What do you think about making this a template? Something like this:


```suggestion
func NewSchemaGVR(g, v, r string) schema.GroupVersionResource {
	return schema.GroupVersionResource{
		Group:    g,
		Version:  v,
		Resource: r,
	}
}
```

extensions/defaults/timeouts/timeouts.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TwentyMinuteWatch() *int64 {
+	timeout := int64(60 * 20)
+	return &timeout
+}
+
+func ThirtyMinuteWatch() *int64 {
+	timeout := int64(60 * 30)
+	return &timeout
+}
```
caliskanugur, line 28: Same for this one too, if we add a new timeout duration, isn't this file going to get longer for the same const?

```suggestion
func NewTimeoutDuration(mins int) *int64 {
	timeout := int64(60 * mins)
	return &timeout
}
```

extensions/provisioninginput/machinepools.yml
```diff
@@ -0,0 +1,38 @@
+etcdControlPlaneWorker:
```
caliskanugur, line 1: Where these YAML files are used? Are these going to be used on r/r side?

caliskanugur (commented):
Looks good, I've got couple of questions

extensions/defaults/timeouts/timeouts.go
susesgartner: 1. I like the idea your proposing so i'll implement it
2. Would it make sense for me to create a second constant in addition to NewTimeoutDuration that is the default that we use across the majority of our framework. The reason I mention this is because the majority of all of our timeoutDurations are the ThirtyMinuteWatch, and one of the major selling points of a defaults file is to allow us to update the constant in all files without having to go and update each file individually i.e update the input of NewTimeoutDuration in all files.

extensions/provisioninginput/machinepools.yml
susesgartner: The ymls aren't going to be used on the rancher side but the functions that pull the data from the ymls will be used.

extensions/defaults/timeouts/timeouts.go
caliskanugur: For sure that makes sense to me, I agree, for your second point; Do you mean something like this:

```
type Duration int

const (
  ThirtyMinutes Duration = 30
  FifteenMinutes Duration = 15
)

func NewTimeoutDuration(mins Duration) *int64 {
	timeout := int64(60 * int(mins))
	return &timeout
}
```

extensions/provisioninginput/machinepools.yml
caliskanugur: Could you elaborate on this one for me, I don't think I understand this move yet, is there a code reference to these files that I am potentially missing?

extensions/defaults/timeouts/timeouts.go
susesgartner: Yep, that would be a good way of handling it. I'll get that implemented.

extensions/provisioninginput/machinepools.yml
susesgartner: If you look in [pools.go](https://github.com/rancher/shepherd/blob/713005f72c344fa4afde5cb11d283383aecc261b/extensions/provisioninginput/pools.go) the 2 files are referenced by their respective functions. The main design goal here is to provide a way to store parts of our configs that are widely used and not sensitive without using global vars.

caliskanugur: Oh, I missed it, thanks for the reference. Could you get into more details about the design selection, what's the net positive implementation of having a YAML file, reading that file, and making it a Golang object, instead of using it as a Golang object?

extensions/defaults/schema/groupversionresources/groupversionresources.go
caliskanugur: These would be constructor functions. They all seem like **functions** that return **fixed** values. So one way or another they are functions right now, right? So what I am trying to say is, is it expected to add a new function for every new GVR instead of having a constructor function?

extensions/provisioninginput/machinepools.yml
susesgartner: There are several positives:
1. Cleaner code. We use quite a few objects (which are basically yaml data) throughout the framework where we declare a default i.e machine pools, nodepools, group version resources... etc. When declaring these defaults we currently can declare them in place which is often not very reusable or as easily maintained (especially if its declared in place in multiple locations) or we can declare a global var (bad coding practice). Machine pools is a great example of this, it is widely used across the framework/rancher and we would have quite a large amount of duplicate code if we declared a Golang object in place. 
2. Simpler Configuration: This also opens the door to creating more defaults for our configs. If you look at machinepools there are a finite number of machine pool configurations that we test with, this is especially true in the case of our static tests. The same applies to other parts of provisioning input as well i.e machineConfigs, etcd... etc and while we cannot store all parts of the config as defaults (vpcID, credentials, amis...) we can store a lot of it and reduce the amount of configuration required to create clusters. 

Overall I think that putting the equivalent of yaml data in a .go file is not as reusable, maintainable, or modular as storing it in a yaml file and then loading it wherever you need it (which is a 1 liner). _yaml data in a yaml file, go code in a go file_

igomez06: 2. I don't know if I agree with this at all. If we want to test the different configurations out we would be that in the test itself. 

So I don't have a huge problem with this, but I have two concerns. One it completely just copies over our load config func. 

The other why is only the machine Pools default yaml but not the other defaults not? There is no consistency. The purpose of these is to set the roles themselves, and honestly only for the tests we have. I feel like this shouldn't even be in shepherd but in the provisioning directory in rancher/rancher instead. These are defaults for our specific tests.

igomez06: After we get our actions vs extensions resolved. This will most likely move anyways

susesgartner: I agree that test specific configurations belong in the test they are designed for, but when a configuration is used in a variety of tests I generally lean towards putting it in a place were it can be reused instead of redeclared in multiple tests. 

The load config function, while similar in the sense that it loads data into an object, is not the same as the LoadConfig function which points specifically at the cattle_test_config and uses the github.com/cresty/defaults package.

I don't know if I follow what you mean by "no consistency". Within the defaults directory it is consistent, anything that is widely used across rancher/shepherd that can be stored as a const is stored in the defaults dir and when we require const objects we load them from a yaml file. The only exception to this rule is the groupversionresources and that only occurs because it is an upstream k8s object without struct tags. 

For the rest of what you have mentioned I would say that **most** of the defaults dir will end up in rancher after actions vs extensions gets resolved.

extensions/defaults/schema/groupversionresources/groupversionresources.go
susesgartner: Yes, they are returning fixed values, they are the equivalent of a "const" @caliskanugur. The 2 concerns I have for axing them and putting a nice generic constructor  are:
1. They would no longer be equivalent in function to a "const". This is the case since it would be the equivalent of declaring each GVR in place wherever it is called. i.e NewSchemaGVR("x", "y", "z") is dynamic, not const and therefore there is no way to update/reuse a given GVR across the framework.
2. If we go with the above approach then I would vote on moving the constructor into the Schema directory since it would no longer be a "const".

extensions/provisioninginput/machinepools.yml
igomez06: Agreed this will get resolved anyways, and I'm honestly fine with this.

igomez06: I do think the LoadDefault function should not be here then, and more in the pkg. What do you think?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/rancher/shepherd
 
-go 1.22
+go 1.22.0
```
igomez06, line 3: why is this change necessary?

extensions/provisioninginput/machinepools.yml
igomez06: The consistency thing I was referring to, is why is this in a yaml where as the GroupVersions in functions. That is what I am talking about. I want to reiterate, my overall concern is that some thing is added to the yaml and we're not testing what we say we're testing. Does that make sense?

go.mod
susesgartner: It is only necessary on macs due to a golang issue. I will remove it.

extensions/provisioninginput/machinepools.yml
susesgartner: Going to reiterate: If I could put the GroupVersions in a yaml I would however, I cannot because they are upstream k8s objects that do not have struct tags and as such I cannot simply unmarshal or change them.

susesgartner: I am not sure I understand your final comment. Can you give me an example of something that of something that gets added to the yaml that would be a problem?

PR #110, opened 2024-03-18
Modify CreateNamespace to accept clusterID and projectName as inputs instead of norman project

Priyashetty17:
Currently, `Create Namespace` function in the kube api package accepts norman project as input. However, for automating test cases for Projects using the public API, I prefer not to reference the norman project or use functions that require norman or steve project input. So, to keep the create namespace function generic and not tied to any specific API (norman/steve/public), I'm modifying the function to accept clusterID and projectName as inputs instead of norman project.

**Backports:**
- 2.7: https://github.com/rancher/shepherd/pull/111
- 2.8: https://github.com/rancher/shepherd/pull/112

### DIFF
--- a/extensions/kubeapi/namespaces/create.go
+++ b/extensions/kubeapi/namespaces/create.go
@@ -3,10 +3,8 @@ package namespaces
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/rancher/shepherd/clients/rancher"
-	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
 	"github.com/rancher/shepherd/extensions/defaults"
 	"github.com/rancher/shepherd/extensions/unstructured"
 	"github.com/rancher/shepherd/pkg/api/scheme"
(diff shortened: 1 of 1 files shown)


extensions/kubeapi/namespaces/create.go
```diff
@@ -21,17 +19,20 @@ import (
 
 // CreateNamespace is a helper function that uses the dynamic client to create a namespace on a project.
 // It registers a delete function with a wait.WatchWait to ensure the namspace is deleted cleanly.
-func CreateNamespace(client *rancher.Client, namespaceName, containerDefaultResourceLimit string, labels, annotations map[string]string, project *management.Project) (*coreV1.Namespace, error) {
-	// Namespace object for a project name space
+func CreateNamespace(client *rancher.Client, clusterID, projectName, namespaceName, containerDefaultResourceLimit string, labels, annotations map[string]string) (*coreV1.Namespace, error) {
```
anupama2501, line 22: can we change this to projectID as its a projectID rather than a name?

Priyashetty17: Initially I had projectID but I later changed it to projectName because there is no project ID in public API. 
```
> k get projects p-bg8wt -n c-m-5w796d5p -o yaml | egrep -h 'id|name'
    lifecycle.cattle.io/create.project-namespace-auth_c-m-5w796d5p: "true"
  - clusterscoped.controller.cattle.io/project-namespace-auth_c-m-5w796d5p
  name: p-bg8wt
  namespace: c-m-5w796d5p
  uid: e839244a-a956-427c-b444-8568d45003a0
> 
```

anupama2501 (approved):
lgtm

igomez06 (approved):
LGTM.

PR #114, opened 2024-03-19
[2.8] Backup and Restore Extensions

nickwsuse:
Extensions in this PR are to enable and enhance backup/restore checks.


nickwsuse:
Charts extensions were moved into the actions/extensions part of r/r so this PR is no longer needed.

PR #117, opened 2024-03-22
Update Rancher leader extension to use lease instead of configmap

Priyashetty17:
## Issue: 
https://github.com/rancher/qa-tasks/issues/1239

## Problem: 
- Starting from Rancher version **2.8.3** and continuing into version 2.9.X, Rancher now uses `only Lease objects `for leader election. Before this change, it used a mix of ConfigMap and Lease objects. This implies that `lease object exists in 2.7, 2.8 and 2.9 releases`. So, "rancherleader" extension should be updated to use lease object instead of configmap.

## Solution
- Update the "rancherleader" extension  to retrieve the rancher leader pod name using lease object.
```
https://Rancher_url/v1/coordination.k8s.io.leases?fieldSelector=metadata.name=cattle-controllers
```

**Backports:** Since `Lease` object is available in 2.7 and 2.8, we need backports for this
- 2.8: https://github.com/rancher/shepherd/pull/124
- 2.7: https://github.com/rancher/shepherd/pull/123

**Note**: Marking this blocker since without this PR, the existing tests using this extension won't work for 2.8.3 and 2.9.

### DIFF
--- a/extensions/rancherleader/rancherleader.go
+++ b/extensions/rancherleader/rancherleader.go
@@ -1,40 +1,33 @@
 package rancherleader
 
 import (
-	"encoding/json"
-
 	"github.com/rancher/shepherd/clients/rancher"
-	"github.com/rancher/shepherd/extensions/configmaps"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
+	coordinationv1 "k8s.io/api/coordination/v1"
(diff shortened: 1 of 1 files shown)


caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

PR #120, opened 2024-03-22
[2.8] Add a Workflow That Verifies Changes

caliskanugur:
Adds a workflow that:

- Verifies mod file changes
- Builds packages inside `pkg`, `client`, and `extensions` dirs
- Uses `golangci-lint` to check linter issues

Triggers on:

- Every push (merges in this case)
- And when a pull request opens targeting `release/*` or `main`

### DIFF
--- a/.github/scripts/build-packages.sh
+++ b/.github/scripts/build-packages.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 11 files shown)


floatingman (approved):

igomez06 (approved):
LGTM

PR #127, opened 2024-03-25
Add Wrangler client to shepherd with cleanup functions

floatingman:
To use Wrangler in Rancher automation tests, it is necessary to import the associated Wrangler functions and override methods to provide access to the test session for test cleanup purposes


floatingman:
Major additions are: 
clients/wrangler/clients/clients.go

PR #133, opened 2024-03-27
Added Wrangler client with cleanup and Wrangler Context

floatingman:
To use Wrangler for various functions in Shepherd, it is necessary to create a client and context used by Wrangler so cleanup will be performed in tests.

I know many files are included in this PR, but most are just from generated code.

Some files of note to look at are:
clients/wrangler/clients/clients.go
pkg/wrangler/context.go
pkg/schemas/management.cattle.io/v3/schema.go


floatingman:
https://github.com/rancher/qa-tasks/issues/1179

PR #145, opened 2024-04-08
[v2.9] Convert flaky Python CLI tests to Go

markusewalker:
### PR Description
As part of broader efforts to fix flaky tests, it was discussed to shift away from the flaky Python tests for the Rancher CLI and transition them over to Go. As such, all of the tests minus apps have been translated. This is because `rancher apps` has known failures in the Rancher CLI.

### DIFF
--- a/clients/rancher/client.go
+++ b/clients/rancher/client.go
@@ -17,6 +17,7 @@ import (
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 
 	kubeProvisioning "github.com/rancher/shepherd/clients/provisioning"
+	"github.com/rancher/shepherd/clients/ranchercli"
 	kubeRKE "github.com/rancher/shepherd/clients/rke"
 	"github.com/rancher/shepherd/pkg/clientbase"
 	"github.com/rancher/shepherd/pkg/config"
@@ -41,6 +42,8 @@ type Client struct {
 	Catalog *catalog.Client
 	// Config used to test against a rancher instance
 	RancherConfig *Config
+	// CLI is the client used to interact with the Rancher CLI
+	CLI *ranchercli.Client
 	// Session is the session object used by the client to track all the resources being created by the client.
 	Session *session.Session
 	// Flags is the environment flags used by the client to test selectively against a rancher instance.
@@ -85,6 +88,13 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 	}
 
 	c.Steve.Ops.Session = session
+
+	if rancherConfig.RancherCLI {
+		c.CLI, err = ranchercli.NewClient(session, bearerToken, rancherConfig.Host, c.Management)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	catalogClient, err := catalog.NewForConfig(restConfig, session)
 	if err != nil {
--- a/clients/ranchercli/client.go
+++ b/clients/ranchercli/client.go
@@ -1,0 +1,95 @@
+package ranchercli
+
+import (
+	"fmt"
+	"os/exec"
+
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	namegen "github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/shepherd/pkg/session"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	rancher   = "rancher"
+	createCmd = "create"
+	deleteCmd = "delete"
+	moveCmd   = "move"
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	args = append([]string{name, createCmd}, args...)
+	err := c.ExecuteCommand(rancher, args...)
+	if err != nil {
+		return err
+	}
+
+	c.Session.RegisterCleanupFunc(func() error {
+		return c.Delete(name, args...)
+	})
+
+	return nil
+}
+
+func (c *Client) Delete(name string, args ...string) error {
+	args = append([]string{name, deleteCmd}, args...)
+
+	return c.ExecuteCommand(rancher, args...)
+}
+
+func (c *Client) Exists(rancher, resourceType, name string) error {
+	cmdStr := fmt.Sprintf("%s %s ls | grep %s", rancher, resourceType, name)
+	cmd := exec.Command("sh", "-c", cmdStr)
+	err := cmd.Run()
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() != 0 {
+				return fmt.Errorf("error validating %s exists: %v", name, err)
+			}
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+// ExecuteCommand will execute a command and log any errors.
+func (c *Client) ExecuteCommand(name string, args ...string) error {
+	command := exec.Command(name, args...)
+	_, err := command.Output()
+	if err != nil {
+		return fmt.Errorf("error executing command '%s': %v", name, err)
+	}
+
+	return nil
+}
+
+// NewClient will download the CLI and login to the Rancher server.
+func NewClient(session *session.Session, token, host string, client *management.Client) (*Client, error) {
+	c := &Client{
+		Session: session,
+	}
+
+	projectConfig := &management.Project{
+		ClusterID: "local",
+		Name:      namegen.AppendRandomString("project"),
+	}
+
+	testProject, err := client.Project.Create(projectConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Logging into Rancher server...")
+	err = c.ExecuteCommand(rancher, "login", "--token", token, "https://"+host, "--skip-verify", "--context", testProject.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
--- a/extensions/cli/cli.go
+++ b/extensions/cli/cli.go
@@ -1,0 +1,139 @@
+package cli
+
+import (
+	ranchercli "github.com/rancher/shepherd/clients/ranchercli"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	rancher    = "rancher"
+	context    = "context"
+	projects   = "projects"
+	namespaces = "namespaces"
+	catalog    = "catalog"
+)
+
+// SwitchContext will display the current context and switch to the default one.
+func SwitchContext(client *ranchercli.Client) error {
+	logrus.Infof("Listing the current context...")
+	err := client.ExecuteCommand(rancher, context, "current")
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Switching to the default context...")
+	err = client.ExecuteCommand(rancher, context, "switch")
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateProjects will create and projects in the specified cluster.
+func CreateProjects(client *ranchercli.Client, projectName, cluster string) error {
+	logrus.Infof("Creating a project...")
+	err := client.ExecuteCommand(rancher, projects, "create", "--cluster", cluster, projectName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Validating the project exists...")
+	err = client.Exists(rancher, projects, projectName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// DeleteProjects will delete projects in the specified cluster.
+func DeleteProjects(client *ranchercli.Client, projectName string) error {
+	logrus.Infof("Deleting the project...")
+	err := client.Delete(projects, projectName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Validating the project is deleted...")
+	err = client.ExecuteCommand(rancher, projects, "ls", "|", "grep", projectName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateNamespaces will create namespaces in the specified cluster.
+func CreateNamespaces(client *ranchercli.Client, namespaceName, projectName string) error {
+	logrus.Infof("Creating a namespace in default project...")
+	err := client.Create(namespaces, namespaceName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Validating the namespace exists...")
+	err = client.Exists(rancher, namespaces, namespaceName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Creating test project...")
+	err = client.Create(projects, "--cluster", "local", projectName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Validating the project exists...")
+	err = client.Exists(rancher, projects, projectName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// DeleteNamespaces will delete namespaces in the specified cluster.
+func DeleteNamespaces(client *ranchercli.Client, namespaceName, projectName string) error {
+	logrus.Infof("Deleting the namespace...")
+	err := client.Delete(namespaces, namespaceName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Deleting the project...")
+	err = client.Delete(projects, projectName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateCatalogs will create and delete catalogs in the specified cluster.
+func CreateCatalogs(client *ranchercli.Client, catalogName string) error {
+	logrus.Infof("Creating a catalog...")
+	err := client.ExecuteCommand(rancher, catalog, "add", catalogName, "https://git.rancher.io/helm3-charts")
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Validating the catalog exists...")
+	err = client.Exists(rancher, catalog, catalogName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// DeleteCatalogs will delete catalogs in the specified cluster.
+func DeleteCatalogs(client *ranchercli.Client, catalogName string) error {
+	logrus.Infof("Deleting the catalog...")
+	err := client.Delete(catalog, catalogName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/clients/rancher/config.go
+++ b/clients/rancher/config.go
@@ -14,4 +14,5 @@ type Config struct {
 	CACerts       string `yaml:"caCerts" json:"caCerts" default:""`
 	ClusterName   string `yaml:"clusterName" json:"clusterName" default:""`
 	ShellImage    string `yaml:"shellImage" json:"shellImage" default:""`
+	RancherCLI    bool   `yaml:"rancherCLI" json:"rancherCLI" default:"false"`
 }


extensions/cli/cli.go
```diff
@@ -0,0 +1,91 @@
+package cli
```
caliskanugur, line 1: This looks good. Markus, could you check helm and rkecli clients, do you think this should be another client? I think we could have this as a client. And these functions as extensions still, but they would depend on that client. What do you think?

caliskanugur (commented):
Looks good, I've got one question

extensions/cli/cli.go
markusewalker: I'm unsure what part would be a client? Are you saying you think just everything in this shepherd PR should be in a separate client package instead?

markusewalker: Resolved

extensions/cli/cli.go
```diff
@@ -0,0 +1,109 @@
...
+
+	logrus.Infof("Creating a project...")
+	err := client.Create("./rancher", "projects", "create", "--cluster", "local", "auto-cli-project")
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Deleting the project...")
+	err = client.Create("./rancher", "projects", "delete", "auto-cli-project")
+	if err != nil {
```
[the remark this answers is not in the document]
markusewalker: The idea behind the Rancher CLI is that it's an alternative method to do calls outside of Norman, Steve, etc. So users that use Rancher CLI wouldn't mix and match. So for rancher projects, it would just be the command `rancher projects ls` to see if what we just made works.

From the points above, that isn't going to give what you want, but that is how a CLI user would validate their projects are existing.

extensions/cli/cli.go
```diff
@@ -0,0 +1,109 @@
...
+	err := client.Create("./rancher", "namespaces", "create", "auto-cli-ns")
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Creating test project...")
+	err = client.Create("./rancher", "projects", "create", "--cluster", "local", "auto-cli-project")
+	if err != nil {
+		return err
+	}
```
slickwarren, line 67: we talked offline a little about this but I think it would be worth adding a negative test case here too. I.e. this line could be a good candidate, try to create a project in a non-existing cluster and make sure that the client doesn't report a false positive.

clients/rancher/client.go
```diff
@@ -41,6 +42,8 @@ type Client struct {
 	Catalog *catalog.Client
 	// Config used to test against a rancher instance
 	RancherConfig *Config
+	// RancherCLI is the client used to interact with the Rancher CLI
```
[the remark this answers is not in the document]
markusewalker: I'll fix it, that was just an oversight in previous drafts.

extensions/cli/cli.go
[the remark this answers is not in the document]
slickwarren: I am okay with using the CLI to list what is created, as long as there's some validation after the creation call I'm good -- I do think steve would be the proper way to QA it but I can see your point too so whatever you're comfortable adding

extensions/cli/cli.go
markusewalker: Implemented an Exists function that should address this.

extensions/cli/cli.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// CreateAndDeleteNamespaces will create and delete namespaces in the local cluster.
+func CreateAndDeleteNamespaces(session *session.Session) error {
+	client := &ranchercli.Client{
+		Session: session,
+	}
+
+	logrus.Infof("Creating a namespace in default project...")
+	err := client.Create("./rancher", "namespaces", "create", "auto-cli-ns")
```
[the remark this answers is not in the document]
markusewalker: Resolved.

clients/ranchercli/download.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+func (c *Client) Exists(rancher, resourceType, name string) (bool, error) {
+	cmdStr := fmt.Sprintf("%s %s ls | grep %s", rancher, resourceType, name)
+	cmd := exec.Command("sh", "-c", cmdStr)
+	err := cmd.Run()
+	if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == 1 {
```
slickwarren, line 29: is there a reason this is only for `1`? If not, can you switch it to `if exitError.ExitCode() != 0`?

clients/ranchercli/download.go
```diff
@@ -0,0 +1,96 @@
...
+func (c *Client) Create(name string, args ...string) error {
+	c.ExecuteCommand(name, args...)
+
+	return nil
+}
+
+func (c *Client) Exists(rancher, resourceType, name string) (bool, error) {
+	cmdStr := fmt.Sprintf("%s %s ls | grep %s", rancher, resourceType, name)
+	cmd := exec.Command("sh", "-c", cmdStr)
+	err := cmd.Run()
```
slickwarren, line 26: I don't think this is the error we would want. Even if `name` isn't in the output of ls, the grep command will not exit with a nonzero code afaik. 

Instead I think you'd have to get the output of the `ls` or write a short bash if statement that will fail if the `name` isn't in the output of ls. maybe `if [[ -n $(%s %s ls | grep %s) ]]; then exit(0); else; exit(1); fi;` ?

slickwarren (changes requested):

clients/ranchercli/download.go
markusewalker: The `return false, err` on line 35 is capturing other exit codes. Bash commands typically are going to give exit code 1 after running `echo $?` when returning an empty result.

clients/ranchercli/download.go
markusewalker: I actually tested this in the separate functions with bogus names  like `test` and `bogus` and it errored each time.

clients/ranchercli/download.go
igomez06: I don't get this logic you're still returning false on line 35 why is 1 so special?

extensions/cli/cli.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	rancher    = "./rancher"
+	context    = "context"
+	projects   = "projects"
+	namespaces = "namespaces"
+	catalog    = "catalog"
+)
+
+// SwitchContext will display the current context and switch to the default one.
+func SwitchContext(session *session.Session) error {
```
igomez06, line 19: Why are these extensions, and not member functions of the client?

igomez06: Once you answer that it would make sense why you create new clients for each one.

clients/ranchercli/download.go
markusewalker: I mean, `1` just for reasons above. But if both of you think it should just be switched to `!= 0`, it's not a huge deal to me to change it.

extensions/cli/cli.go
markusewalker: When I initially talked to Ugur on transitioning this over to clients, the initial code had `downloads.go` not in the format of a client. When we discussed, at the time it made more sense to keep the `cli.go` as extensions still and to call the `Create` and `Exists` functions.

clients/ranchercli/download.go
igomez06: this is to simplify logic 3 embedded if statements is hard to read.

extensions/cli/cli.go
igomez06: But you're creating new clients each time which doesn't makes sense. I would like to hear what Ugur says, but these seem that they need to be member functions for the cli client. Reinitializing the client each time is a no go

igomez06: Or at least have it take a cli client as a parameter. I will wait to see what Ugur says as well.

clients/ranchercli/download.go
igomez06: I understand now an exit code of zero shouldn't return false that's why you're checking. I think !=0 would be best. Thanks!

extensions/cli/cli.go
markusewalker: Maybe we can talk offline of this. I'm still not fully knowledgeable of clients and best practices, so it would be good to get up to speed on what to do versus what not to do.

clients/ranchercli/download.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	c.ExecuteCommand(name, args...)
+
+	return nil
```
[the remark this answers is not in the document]
markusewalker: Resolved

clients/ranchercli/download.go
markusewalker: Resolved

extensions/cli/cli.go
markusewalker: I believe this may be resolved now? Planning to discuss offline to verify if this is the case.

clients/ranchercli/download.go
slickwarren: hmm. mine doesn't error on a linux machine 🤔

clients/ranchercli/download.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	err := c.ExecuteCommand(name, args...)
+	if err != nil {
```
igomez06, line 19: You don't need this if statement just return  c.ExecuteCommand(name, args...)

clients/ranchercli/download.go
markusewalker: Met offline and got this working.

clients/ranchercli/download.go
markusewalker: Resolved

clients/ranchercli/download.go
```diff
@@ -0,0 +1,92 @@
+package ranchercli
+
+import (
```
igomez06, line 3: Hey Markus why is this called download.go?

markusewalker: It was initially because it was just setting up the Rancher CLI, but it's a bit outdated now. It can be renamed to something else, I don't have a preference as to exact name

markusewalker: I'll rename to `client.go`.

extensions/cli/cli.go
```diff
@@ -0,0 +1,147 @@
...
+
+// CreateAndDeleteCatalogs will create and delete catalogs in the local cluster.
+func CreateAndDeleteCatalogs(client *ranchercli.Client) error {
+	var catalogName = namegen.AppendRandomString("catalog")
+
+	logrus.Infof("Creating a catalog...")
+	err := client.Create(rancher, catalog, "add", catalogName, "https://git.rancher.io/helm3-charts")
+	if err != nil {
+		return err
+	}
```
caliskanugur, line 132:
```suggestion
	err := client.Executecommand("I have", "another", "action", catalogName, "https://git.rancher.io/helm3-charts")
	if err != nil {
		return err
	}
```

clients/ranchercli/client.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	return c.ExecuteCommand(name, args...)
+}
```
caliskanugur, line 19:
```suggestion

rancher := "rancher"

func (c *Client) Create(name string, args ...string) error {

       firstPiece := args[0-1]
       secondPiece := args[1:]
 c.ExecuteCommand(rancher, name, firsPiece, "create", secondPiece)
 
 c.session.RegisterCleanup(
 c.Delete(args )
 )

}

func (c *Client) Delete(name string, args ...string) error {
       firstPiece := args[0-1]
       secondPiece := args[1:]
 return c.ExecuteCommand(rancher, name, firsPiece, "delete", secondPiece)
}

func (c *Client) Get(name string, args ...string) error {
	return c.ExecuteCommand(name, args...)
}

func (c *Client) List(name string, args ...string) error {
	return c.ExecuteCommand(name, args...)
}

func (c *Client) Update(name string, args ...string) error {
	return c.ExecuteCommand(name, args...)
}

yadya := newClient
yadaya.CLI.Create(projects, "create", "--cluster", "local", projectName)


```

caliskanugur (commented):
Offline synched items with @markusewalker

clients/ranchercli/client.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/rancher/shepherd/pkg/session"
+	"github.com/sirupsen/logrus"
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	return c.ExecuteCommand(name, args...)
```
anupama2501, line 18: why do we want to have the Create function calling the ExecuteCommand function? Can we not have just the ExecuteCommand?

extensions/cli/cli.go
```diff
@@ -0,0 +1,147 @@
...
+
+	logrus.Infof("Validating the project exists...")
+	err = client.Exists(rancher, projects, projectName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Deleting the project...")
+	err = client.Create(rancher, projects, "delete", projectName)
```
anupama2501, line 54: Can we add a check here to make sure when we list the projects, the deleted project wont show up?

clients/ranchercli/client.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return nil
+}
+
+// NewClient will download the CLI and login to the Rancher server.
+func NewClient(version, token, host string) (*Client, error) {
+	client := &Client{
+		Session: &session.Session{},
+	}
```
caliskanugur, line 53:
```suggestion
func NewClient(session *session.Session, version, token, host string) (*Client, error) {
	client := &Client{
		Session: session,
	}
```

clients/rancher/client.go
```diff
@@ -86,6 +89,15 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 
 	c.Steve.Ops.Session = session
 
+	if rancherConfig.CLIVersion != "" {
+		c.CLI, err = ranchercli.NewClient(rancherConfig.CLIVersion, bearerToken, rancherConfig.Host)
+		if err != nil {
+			return nil, err
+		}
+
+		c.CLI.Session = session
```
caliskanugur, line 98:
```suggestion
		c.CLI, err = ranchercli.NewClient(session, rancherConfig.CLIVersion, bearerToken, rancherConfig.Host)
		if err != nil {
			return nil, err
		}
```

caliskanugur (commented):
I've got one last comment here, I feel like this way the session should be in use with the create call. What do you think?

anupama2501 (commented):
Have a few questions and comments:

Can we separate out DeleteProject/CreateProject/CreateNamespace as individual functions and then call these in the respective tests? That way it will be easier to use these extensions, outside of the P0 tests for cli?

extensions/cli/cli.go
markusewalker: Resolved

clients/ranchercli/client.go
markusewalker: Resolved

clients/rancher/client.go
markusewalker: Resolved

clients/ranchercli/client.go
markusewalker: Resolved

clients/ranchercli/client.go
markusewalker: Resolved, spoke offline regarding this.

extensions/cli/cli.go
markusewalker: Resolved, spoke offline regarding this.

extensions/cli/cli.go
```diff
@@ -0,0 +1,147 @@
...
+	err = client.Create(rancher, context, "switch")
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateAndDeleteProjects will create and delete projects in the local cluster.
+func CreateAndDeleteProjects(client *ranchercli.Client) error {
```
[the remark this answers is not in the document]
markusewalker: Resolved, spoke offline regarding this.

extensions/cli/cli.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	logrus.Infof("Deleting the project...")
+	err = client.Create(rancher, projects, "delete", projectName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Creating project in non-existent cluster...")
+	err = client.Create(rancher, projects, "create", "--cluster", missingCluster, projectName)
```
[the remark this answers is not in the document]
markusewalker: Resolved, spoke offline regarding this.

extensions/cli/cli.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	logrus.Infof("Validating the catalog exists...")
+	err = client.Exists(rancher, catalog, catalogName)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Deleting the catalog...")
+	err = client.Create(rancher, catalog, "delete", catalogName)
```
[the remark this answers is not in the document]
markusewalker: Resolved, spoke offline regarding this.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+	})
+
+	return nil
+}
+
+func (c *Client) Delete(name string, args ...string) error {
+	args = append([]string{name, deleteCmd}, args...)
+	c.ExecuteCommand(rancher, args...)
+
+	return nil
```
igomez06, line 39: Markus you have to do `return c.ExecuteCommand(rancher, args...)` we're completely ignoring the error returned.

clients/rancher/client.go
```diff
@@ -86,6 +90,25 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 
 	c.Steve.Ops.Session = session
 
+	if rancherConfig.CLIVersion != "" {
+		projectConfig := &management.Project{
+			ClusterID: "local",
+			Name:      namegen.AppendRandomString("project"),
+		}
+
+		testProject, err := c.Management.Project.Create(projectConfig)
```
igomez06, line 99: Why are we creating a project? Creating a resource here seems off.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, err
+	}
+
+	logrus.Infof("Logging into Rancher server...")
+	err = client.ExecuteCommand("./rancher", "login", "--token", token, "https://"+host, "--skip-verify", "--context", projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{}, nil
```
igomez06, line 111: Why are you returning an empty struct?

clients/rancher/client.go
```diff
@@ -86,6 +90,25 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		c.CLI, err = ranchercli.NewClient(session, rancherConfig.CLIVersion, bearerToken, rancherConfig.Host, testProject.ID)
+		if err != nil {
+			return nil, err
+		}
+
+		c.CLI.Session = session
```
igomez06, line 109: If you pass the session you don't need to do this Markus

igomez06 (commented):
it's almost there gave more feedback and more questions.

clients/rancher/client.go
markusewalker: The Rancher CLI `login` command by default uses an interactive method to choose which project you want to login to. To bypass this and login silently, you need to provide the project ID. That's why we have this here so that we're able to bypass this issue in the CLI.

clients/ranchercli/client.go
markusewalker: I'll fix that.

clients/rancher/client.go
markusewalker: Heard, I was following the examples above, but that's a good point. I'll remove it

clients/ranchercli/client.go
markusewalker: That was from a prior iteration, I'll return the client. Good catch!

clients/ranchercli/client.go
markusewalker: Resolved.

clients/ranchercli/client.go
markusewalker: Resolved.

clients/rancher/client.go
markusewalker: Resolved.

caliskanugur (commented):
Mostly questions and a few comments

clients/rancher/client.go
```diff
@@ -86,6 +90,25 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
...
+	if rancherConfig.CLIVersion != "" {
+		projectConfig := &management.Project{
+			ClusterID: "local",
+			Name:      namegen.AppendRandomString("project"),
+		}
+
+		testProject, err := c.Management.Project.Create(projectConfig)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
markusewalker: This is necessary. We need to be able to bypass the interactive option of selecting a project ID. Without doing this, the test is going to fail as you need to interactively select the cluster you are going to login to.

extensions/cli/cli.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// CreateProjects will create and projects in the specified cluster.
+func CreateProjects(client *ranchercli.Client, projectName, cluster string) error {
+	logrus.Infof("Creating a project...")
+	err := client.ExecuteCommand(rancher, projects, "create", "--cluster", cluster, projectName)
```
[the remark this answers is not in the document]
markusewalker: We talked offline about some of Rancher CLI having nunances in their commands. For `rancher projects`, that is an example where the `rancher projects create` command is more involved then other operations. `rancher namespaces` is a simple example where we don't have to do this as an example is just `rancher namespace create dummy-namespace`.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+	err = osys.Chdir("rancher-" + version)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Logging into Rancher server...")
+	err = client.ExecuteCommand("./rancher", "login", "--token", token, "https://"+host, "--skip-verify", "--context", projectID)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
markusewalker: I initially had it going to `/usr/local/bin`. The issue with that is that you need to have a `sudo` command to do that. In Jenkins, that isn't necessarily a problem, but locally, you have to keep providing a password. Both environments should work the same IMO.

I tried multiple ways to get both to play nice, but putting this approach seemed to be only way that both locally and in Jenkins it works 100% of the time.

If it's already installed, it shouldn't be a problem as it's specifically going to the folder to run the `rancher` binary.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+		return err
+	}
+
+	return nil
+}
+
+// ExecuteCommand will execute a command and log any errors.
+func (c *Client) ExecuteCommand(name string, args ...string) error {
+	command := exec.Command(name, args...)
```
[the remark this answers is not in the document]
markusewalker: I tried doing that and it caused quite a bit of complications in the test cases. I found doing this just kept it simple and have the test cases passing.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+	osys "os"
+	"os/exec"
+	"strings"
+
+	"github.com/rancher/shepherd/pkg/session"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	rancher   = "./rancher"
```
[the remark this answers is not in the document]
markusewalker: I think my other answer above addresses the concern outlined here.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,112 @@
...
+	moveCmd   = "move"
+)
+
+type Client struct {
+	Session *session.Session
+}
+
+func (c *Client) Create(name string, args ...string) error {
+	args = append([]string{name, createCmd}, args...)
+	c.ExecuteCommand(rancher, args...)
```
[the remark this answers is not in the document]
markusewalker: Isn't it just the `Create` that needs it? `Delete` just returns the function and the actual `ExecuteCommand` handles error checking.

clients/ranchercli/client.go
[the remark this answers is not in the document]
igomez06: I have to agree with Ugur. Why couldn't we assume it's been installed? That's what we do for corral as well, our client doesn't install it.

clients/rancher/client.go
igomez06: 1. I don't think this belongs here it should be in the cli client itself.
2. Have a resource creation is more than I would like. Couldn't we use the default project?

markusewalker: I can try and move this over to `cli.go`. But the resource creation is something I don't think we get around, regardless where it is. Especially with downstream clusters being a potential enhancement. 

The Rancher CLI gives options default and system projects. If you only have the local cluster, that could work, but that isn't realistic to what a user would be doing.

markusewalker: Moved the project creation over to `cli.go`.

clients/ranchercli/client.go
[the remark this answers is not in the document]
markusewalker: Briefly touched base with Israel offline. Going to try simplifying this given Ugur's suggestions. Will follow here with results once tested.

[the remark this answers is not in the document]
markusewalker: I was able to get this to work. I'll share examples offline, but going to mark as resolved as this is now transferred over to Jenkins, simplifying this.

clients/ranchercli/client.go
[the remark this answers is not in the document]
markusewalker: This is now being handled in Jenkins, so this has been resolved.

clients/rancher/client.go
markusewalker: Talked offline of this in a meeting. Agreed to leave this in

clients/rancher/client.go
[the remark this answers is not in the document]
markusewalker: Talked offline in a meeting of this, resolving comment

igomez06 (approved):
LGTM
This needs backports right?

markusewalker:
@igomez06 correct. I'll get the needed approvals for the rancher/rancher PRs then open the backports here.

clients/ranchercli/client.go
```diff
@@ -0,0 +1,95 @@
...
+
+// NewClient will download the CLI and login to the Rancher server.
+func NewClient(session *session.Session, token, host string, client *management.Client) (*Client, error) {
+	c := &Client{
+		Session: session,
+	}
+
+	projectConfig := &management.Project{
+		ClusterID: "local",
+		Name:      namegen.AppendRandomString("project"),
```
slickwarren, line 80: I think this was somewhat discussed in the rancher/rancher PR but, confirming we want this hardcoded for now?

slickwarren (approved):

clients/ranchercli/client.go
markusewalker: That is correct.

anupama2501 (approved):
lgtm

caliskanugur (approved):
Looks good to me

PR #148, opened 2024-04-10
Updates/import wrangler main

floatingman:


go.mod
```diff
@@ -29,24 +29,26 @@ require (
 	github.com/rancher/eks-operator v1.3.0
 	github.com/rancher/fleet/pkg/apis v0.0.0-20231017140638-93432f288e79
 	github.com/rancher/gke-operator v1.2.0
-	github.com/rancher/lasso v0.0.0-20230830164424-d684fdeb6f29
+	github.com/rancher/lasso v0.0.0-20240123150939-7055397d6dfa
 	github.com/rancher/norman v0.0.0-20230831160711-5de27f66385d
+	github.com/rancher/rancher v2.2.13+incompatible
```
igomez06, line 34: We can't import rancher/rancher this will cause an import issue.

pkg/wrangler/context.go
```diff
@@ -0,0 +1,343 @@
...
+	metav1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
+	utilruntime.Must(AddToScheme(Scheme))
+	utilruntime.Must(schemes.AddToScheme(Scheme))
+}
+
+type Context struct {
+	RESTConfig *rest.Config
+
+	Apply               apply.Apply
+	Dynamic             *dynamic.Controller
```
igomez06, line 69: We don't need dynamic here. We already have our own client so we don't need this.

clients/wrangler/clients/clients.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/rancher/wrangler/v2/pkg/ratelimit"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type Clients struct {
+	*wrangler.Context
+	Dynamic dynamic.Interface
```
igomez06, line 18: We don't need this here. We have our own

pkg/wrangler/context.go
```diff
@@ -0,0 +1,343 @@
...
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/rancher/lasso/pkg/controller"
+	"github.com/rancher/lasso/pkg/dynamic"
+	"github.com/rancher/norman/types"
+	managementv3api "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/rancher/pkg/settings"
```
igomez06, line 19: we have our own pkg/settings

pkg/wrangler/context.go
```diff
@@ -0,0 +1,343 @@
...
+
+type Context struct {
+	RESTConfig *rest.Config
+
+	Apply               apply.Apply
+	Dynamic             *dynamic.Controller
+	Mgmt                managementv3.Interface
+	ControllerFactory   controller.SharedControllerFactory
+	MultiClusterManager MultiClusterManager
+	K8s                 *kubernetes.Clientset
```
igomez06, line 73: Don't this just yet either. Since this won't have our session clean up.

pkg/wrangler/context.go
floatingman: Do we have the necessary functions in that pkg/settings?   Are you referring to extensions/settings?

pkg/wrangler/context.go
floatingman: This is used by several functions in context.go

pkg/codegen/main.go
```diff
@@ -9,25 +9,40 @@ import (
...
+	err := os.Unsetenv("GOPATH")
+	if err != nil {
+		return
+	}
 
 	controllergen.Run(args.Options{
 		OutputPackage: "github.com/rancher/shepherd/pkg/generated",
 		Boilerplate:   "pkg/codegen/boilerplate.go.txt",
 		Groups: map[string]args.Group{
+			// Commenting this out to avoid generating the management.cattle.io/v3 controllers
```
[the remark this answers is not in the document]
floatingman: We have to be careful when generating code. I'll add information to the README.

pkg/wrangler/context.go
igomez06: Clarified offline

clients/wrangler/clients/clients.go
floatingman: This is just for setting up the connection between Rancher and the Shepherd. This client is not performing a test cleanup. It is all handled by the controllers

igomez06: This is used a completely separate client, I promise you it doesn't need to be here

pkg/wrangler/context.go
igomez06: Those functions we don't need, and are outside the scope of this PR. We only care about the controllers. Does that make sense? I want to be purposeful on what we add.

clients/wrangler/clients/clients.go
floatingman: It's true!

pkg/wrangler/context.go
floatingman: Removed

pkg/wrangler/context.go
```diff
@@ -0,0 +1,304 @@
...
+	if err := w.ControllerFactory.Start(ctx, 50); err != nil {
+		return err
+	}
+	w.leadership.Start(ctx)
+	return nil
+}
+
+// WithAgent returns a shallow copy of the Context that has been configured to use a user agent in its
+// clients that is the given userAgent appended to "rancher-%s-%s".
+func (w *Context) WithAgent(userAgent string) *Context {
```
igomez06, line 147: Do we need this func? Is this something we can add later? That way we can remove the settings file if we don't need it.

igomez06 (commented):
this looks great. Spoke offline about one sanity check. Plus one question.

pkg/wrangler/context.go
igomez06: it's not a deal breaker just a question, otherwise it's an approved.

floatingman: You're right. It's not needed, and I removed it.

caliskanugur (approved):
Looks good to me

igomez06 (approved):
LGTM

igomez06 (approved):
LGTM

caliskanugur (dismissed):
LGTM

igomez06 (approved):
LGTM

caliskanugur (approved):

PR #151, opened 2024-04-11
adds support for deleting init node

Josh-Diamond:
extensions to support: https://github.com/rancher/rancher/pull/44155

Link to previous PR (draft) for additional context around requested changes:
https://github.com/rancher/shepherd/pull/64

### DIFF
--- a/extensions/clusters/clusters.go
+++ b/extensions/clusters/clusters.go
@@ -1252,7 +1252,7 @@ func logClusterInfoWithChanges(clusterID, clusterInfo string, summary summary.Su
 // WatchAndWaitForCluster is function that waits for a cluster to go unactive before checking its active state.
 func WatchAndWaitForCluster(client *rancher.Client, steveID string) error {
 	var clusterResp *v1.SteveAPIObject
-	err := kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
+	err := kwait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaults.TwoMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
 		clusterResp, err = client.Steve.SteveType(ProvisioningSteveResourceType).ByID(steveID)
 		if err != nil {
 			return false, err
@@ -1263,7 +1263,7 @@ func WatchAndWaitForCluster(client *rancher.Client, steveID string) error {
 	if err != nil {
 		return err
 	}
-	logrus.Infof("waiting for cluster to be up.............")
+	logrus.Infof("waiting for cluster to be up...")
 
 	adminClient, err := rancher.NewClient(client.RancherConfig.AdminToken, client.Session)
 	if err != nil {
--- a/extensions/machinepools/machinepools.go
+++ b/extensions/machinepools/machinepools.go
@@ -12,6 +12,7 @@ import (
 	"github.com/rancher/shepherd/clients/rancher"
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/kubeapi/secrets"
 	nodestat "github.com/rancher/shepherd/extensions/nodes"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -20,8 +21,15 @@ import (
 )
 
 const (
-	active = "active"
-	pool   = "pool"
+	active                   = "active"
+	fleetNamespace           = "fleet-default"
+	initNodeLabelKey         = "rke.cattle.io/init-node"
+	local                    = "local"
+	machineNameSteveLabel    = "rke.cattle.io/machine-name"
+	machinePlanSecretType    = "rke.cattle.io/machine-plan"
+	machineSteveResourceType = "cluster.x-k8s.io.machine"
+	pool                     = "pool"
+	True                     = "true"
 
 	nodeRoleListLength = 4
 )
@@ -289,4 +297,30 @@ func MatchRoleToPool(poolRole string, allRoles []Roles) int {
 	}
 	logrus.Warn("unable to match pool to role, likely missing [roles] in machineConfig")
 	return -1
+}
+
+// GetInitMachine accepts a client and clusterID and returns the "init node" machine
+// object for rke2/k3s clusters
+func GetInitMachine(client *rancher.Client, clusterID string) (*v1.SteveAPIObject, error) {
+	logrus.Info("Retrieving secret and identifying machine...")
+	secret, err := secrets.ListSecrets(client, local, fleetNamespace, metav1.ListOptions{
+		LabelSelector: initNodeLabelKey + "=" + True,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// secret.Items[0] will never change when targeting the init node secret,
+	// as the list has been filtered above to grab the single init node secret
+	initNodeMachineName := secret.Items[0].ObjectMeta.Labels[machineNameSteveLabel]
+
+	logrus.Info("Retrieving machine...")
+	initMachine, err := client.Steve.SteveType(machineSteveResourceType).ByID(fleetNamespace + "/" + initNodeMachineName)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Successfully retrieved machine: %s", initNodeMachineName)
+
+	return initMachine, nil
 }
--- a/extensions/steve/steve.go
+++ b/extensions/steve/steve.go
@@ -1,0 +1,38 @@
+package steve
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	notFound = "not found"
+)
+
+// WaitForSteveResourceDeletion accepts a client, steve resource type, and resource ID, then waits for a steve resource to be deleted
+func WaitForSteveResourceDeletion(client *rancher.Client, interval, timeout time.Duration, steveResourceType, steveID string) error {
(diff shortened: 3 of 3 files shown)


extensions/nodes/nodes.go
```diff
@@ -0,0 +1,92 @@
...
+	initNodeMachineName := ""
+
+	logrus.Info("Identifying init node machine plan...")
+	for _, machinePlan := range machinePlans {
+		v, ok := machinePlan.ObjectMeta.Labels[initNodeLabelKey]
+		if ok && v == "true" {
+			logrus.Infof("Found init node machine plan: %s", machinePlan.Name)
+			initNodeMachineName = machinePlan.Labels[machineNameSteveLabel]
+		}
+	}
```
slickwarren, line 49: why not move what's in this for loop to line 36 instead? you could also break in the if statement since you're done looking for initNode.

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// VerifyDeletedMachineRKE2K3S accepts a client and v1.SteveAPIObject and verifies that
+// the machine was successfully removed from the cluster
+func VerifyDeletedMachineRKE2K3S(client *rancher.Client, deletedMachine *v1.SteveAPIObject) error {
+	logrus.Info("Re-fetching machines...")
```
slickwarren, line 76: log seems unnecessary

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,92 @@
...
+
+	machines, err := client.Steve.SteveType(machineSteveResourceType).ListAll(nil)
```
slickwarren, line 78: general question; can't you use the list function here and if list for the deleted machine returns nil, than it is deleted? (Instead of having the for loop below)

extensions/nodes/nodes.go
```diff
@@ -0,0 +1,92 @@
...
+	local                 = "local"
+	machinePlanSecretType = "rke.cattle.io/machine-plan"
+	machineNameSteveLabel = "rke.cattle.io/machine-name"
+)
+
+// GetInitMachine accepts a client and clusterID and returns the "init node" machine
+// object for rke2/k3s clusters
+func GetInitMachine(client *rancher.Client, clusterID string) (*v1.SteveAPIObject, error) {
+	logrus.Info("Retrieving list of secrets...")
+	secretList, err := secrets.ListSecrets(client, local, fleetNamespace, metav1.ListOptions{})
```
slickwarren, line 25: does listOptions have an entry for the `.Type` that you're looking for? i.e. could you just query listOptions for secret.Type=machinePlanSecretType instead of the for loop below?

slickwarren (commented):
some nits/questions

extensions/nodes/nodes.go
susesgartner: Adding on to several of the filtering/list option things Caleb mentioned: we should always (if possible) use the List filters with API calls to simplify our logic and slightly increase performance. You will also probably need to rework your logging statements after this change.

susesgartner (commented):
Looks pretty good, just add some list filtering to replace the for loops and your good to go!

markusewalker (approved):
LGTM

extensions/clusters/clusters.go
```diff
@@ -1252,7 +1252,7 @@ func logClusterInfoWithChanges(clusterID, clusterInfo string, summary summary.Su
 // WatchAndWaitForCluster is function that waits for a cluster to go unactive before checking its active state.
 func WatchAndWaitForCluster(client *rancher.Client, steveID string) error {
 	var clusterResp *v1.SteveAPIObject
-	err := kwait.Poll(500*time.Millisecond, 2*time.Minute, func() (done bool, err error) {
+	err := kwait.PollUntilContextTimeout(context.TODO(), 2*time.Minute, defaults.FifteenMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
```
slickwarren, line 1255: this now only checks every 2 minutes, I think we should keep it to magnitude in seconds. i.e. 1 or 2 second poll

extensions/machinepools/machinepools.go
```diff
@@ -290,3 +298,27 @@ func MatchRoleToPool(poolRole string, allRoles []Roles) int {
...
+func GetInitMachine(client *rancher.Client, clusterID string) (*v1.SteveAPIObject, error) {
+	logrus.Info("Retrieving secret and identifying machine...")
+	secret, err := secrets.ListSecrets(client, local, fleetNamespace, metav1.ListOptions{
+		LabelSelector: initNodeLabelKey + "=" + True,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	initNodeMachineName := secret.Items[0].ObjectMeta.Labels[machineNameSteveLabel]
```
slickwarren, line 313: can you add a comment that this will always be ordered and [0] is exactly what we're expecting / the order won't change due to the label you're looking at?

slickwarren (dismissed):

extensions/steve/steve.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	notFound = "not found"
+)
+
+// DeleteSteveResource accepts a client, resource type and v1.SteveAPIObject and deletes the resource
+func DeleteSteveResource(client *rancher.Client, steveResourceType string, resource *v1.SteveAPIObject) error {
```
igomez06, line 20: What's the point of having this? You're just doing a delete call?

extensions/steve/steve.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if resource != nil {
+		return errors.New("Resource was not removed")
+	}
+
+	return nil
+}
+
+// WaitForSteveResourceDeletion accepts a client, steve resource type, and resource ID, then waits for a steve resource to be deleted
+func WaitForSteveResourceDeletion(client *rancher.Client, interval, timeout time.Duration, steveResourceType, steveID string) error {
```
igomez06, line 52: My concern about this wait is that it's too generic and deleting resources is different depending on what that is.

igomez06:
Josh why did you remove you original extensions for something is generic?

extensions/steve/steve.go
Josh-Diamond: too generic? 🤔 -- could you elaborate on your concern?  When would this not be applicable for awaiting a steve resource to be deleted?  I'm not understanding what you'd like to see in place of this 🙂

igomez06: Deleting a secret is different than deleting a cluster. We would want to create extensions to help validate those differently. Having such a generic function causes confusion for someone  writing a test

igomez06: Have an extensions that waits for a node to be deleted. Why did you decide to go this route than what you had before?

extensions/steve/steve.go
Josh-Diamond: I felt this was a nice enhancement to our extensions.  Anytime we'd like to delete a steve resource, (and this should be applied to all CRUD operations IMO) we can just call this extension, instead of repeating the client call + err != nil, which would make the tests cleaner, more readable and ultimately DRYer code, as we wouldn't be bloating tests with err != nil every time we want to make a client call.  That was my thinking in creating this.

igomez06: you still need to do a err != nil check this function returns error. Why wouldn't that happen?

extensions/steve/steve.go
susesgartner: If this covers a good chunk of the "wait for delete" use cases though wouldn't it make sense to have a generic function and then if you need something more specific you could just create it in place? @igomez06

Josh-Diamond:
@igomez06 I enhanced my extensions as requested by reviewers to be as generic as possible, as it allows the extension to be applicable to more use cases, which is ideal when creating extensions IMO.

extensions/steve/steve.go
Josh-Diamond: Thanks for the clarification Israel!

this function just awaits the resource to be deleted, which would be applicable to secrets too.  Validating the secret was properly deleted should be a separate extension IMO.  This is purely to just await the resource deletion, and I tried to name the function appropriately, as to not cause any confusion.

I went this route as requested by reviewers.  The concern was using v3 vs v1 and the ask was to refactor out all the v3 calls.  This came about as the result of that re-work.  

If I edit this code to be less generic, we will have to add more extensions, extremely similar to each other, for each specific use case, which just seems unnecessary when it could be tackled through one extension.

igomez06: You make a good point though Sam, I'll drop it.

igomez06:
> @igomez06 I enhanced my extensions as requested by reviewers to be as generic as possible, as it allows the extension to be applicable to more use cases, which is ideal when creating extensions IMO.

Sure Josh make sense.

extensions/steve/steve.go
Josh-Diamond: i guess it is very similar, but in the tests, having require.NoError or assert.NoError makes it more readable IMO.  

What do you want me to do about this?  I like it, but if its not getting merged, i'll nuke it 🙂

igomez06: What's the difference of doing
```
require.NoError(client.Delete)
require.NoError(DeleteSteveResource)
```
?
It really does not add much. You other functions make sense, this does not. Remove this and you got my approval.

Josh-Diamond: Screenshot 2024-04-26 at 10 56 58 AM

Josh-Diamond: updated as requested 👍

igomez06 (approved):
LGTM

markusewalker (approved):

igomez06:
@Josh-Diamond there will be backports correct?

Josh-Diamond:
@igomez06 yes, I was awaiting the review process for `v2.9` first, and once approvals were given I plan on drafting up the backports 👍

slickwarren (approved):

PR #157, opened 2024-04-18
[2.8] Add Alerting Drivers Chart Installation, Update Monitoring and Logging Charts to Support Different Provider Installations

caliskanugur:
## Problem

Chart installations only support RKE1 clusters. Ultimately for chart installations on the Airgap environment, the chart tests need to be enhanced to include an Alerting Driver chart too.
 
## Solution

Add Monitoring and Logging charts' payload mappers. Update chart installs option struct to have cluster meta - which is used within these mappers. Add Alerting Driver installation function that supports RKE1/RKE2/K3s installations.


igomez06 (approved):
LGTM

floatingman (approved):

PR #163, opened 2024-04-26
[main] Fixes for agentEnvVars, updates for proxy tests

MSpencer87:
Fixes for custom cluster reg strings with env vars

Added helper for setting corral bastion addresses


slickwarren:
can you link the rancher/rancher PR here for reference to the updated structs?

extensions/clusters/clusters.go
```diff
@@ -305,6 +306,10 @@ func NewRKE1ClusterConfig(clusterName string, client *rancher.Client, clustersCo
 		newConfig.DefaultPodSecurityAdmissionConfigurationTemplateName = clustersConfig.PSACT
 	}
 
+	if clustersConfig.AgentEnvVars != nil {
+		newConfig.AgentEnvVars = clustersConfig.AgentEnvVarsRKE1
+	}
+
```
slickwarren, line 312: this is already set on line 270, do we need the if != nil logic here?

extensions/clusters/clusters.go
```diff
@@ -450,19 +455,26 @@ func NewK3SRKE2ClusterConfig(clusterName, namespace string, clustersConfig *Clus
 		MachinePools:         machinePools,
 	}
 
+	agentEnvVars := []rkev1.EnvVar{}
```
slickwarren, line 458: can't this be omitted, or does setting an initial object match more closely to what rancher does to create a cluster? (thinking of the case where no env vars are set, this was nil before this PR)

slickwarren (commented):
mostly looks solid, just some comments on possibly redundant initializations

markusewalker (approved):
LGTM

extensions/clusters/clusters.go
igomez06: it could be nil right Michael?

extensions/clusters/clusters.go
igomez06: Good question.

slickwarren (approved):

Issue #171, opened 2024-04-30
Add ability for wrangler to interact with downstream clusters similar to the dynamic Client


Priyashetty17:
We currently do not have a way to interact with downstream clusters using wrangler, similar to how the Dynamic client works. https://github.com/rancher/shepherd/blob/351f9be6d0c0fcf61893b46cbc7930c261c0301b/clients/rancher/client.go#L254-L263

For instance, this will list the namespaces in the local cluster: pr.client.WranglerContext.Core.Namespace().List(metav1.ListOptions{}). However, we need a way to list the namespaces in the downstream cluster.

Priyashetty17:
Closing this as this has been implemented.

PR #177, opened 2024-05-06
Wrangler context to interact with downstream clusters

Priyashetty17:
Issue: https://github.com/rancher/shepherd/issues/171

### DIFF
--- a/pkg/wrangler/context.go
+++ b/pkg/wrangler/context.go
@@ -5,6 +5,7 @@ package wrangler
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -245,4 +246,17 @@ func syncOnlyChangedObjects(option controllerContextType) bool {
 		}
 	}
 	return false
+}
+
+// DownStreamClusterWranglerContext creates a wrangler context to interact with a specific cluster.
+func (w *Context) DownStreamClusterWranglerContext(clusterID string) (*Context, error) {
+	restConfig := *w.RESTConfig
+	restConfig.Host = fmt.Sprintf("https://%s/k8s/clusters/%s", w.RESTConfig.Host, clusterID)
+
+	clusterContext, err := NewContext(context.TODO(), &restConfig, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return clusterContext, nil
 }


floatingman (approved):
This looks good.

clients/rancher/client.go
```diff
@@ -262,6 +262,19 @@ func (c *Client) GetDownStreamClusterClient(clusterID string) (dynamic.Interface
 	return dynamic, nil
 }
 
+// GetDownStreamClusterContext is a helper function that creates a wrangler context to interact with a specific cluster.
+func (c *Client) GetDownStreamClusterContext(clusterID string) (*wrangler.Context, error) {
```
igomez06, line 266: I think it would make more sense Priya if we made it like the steve client and do it like this. But same concept.

https://github.com/rancher/shepherd/blob/2bcb20d8aa618d9029dce71939f31d4b32b49dbc/clients/rancher/v1/client.go#L128

igomez06: @floatingman @caliskanugur what do you think?

floatingman: I'm trying to think of why @igomez06? 
Some questions that come to mind are:
Why the ProxyDownstream function wasn't implemented for the latest rancher client?  
Is it more robust?

I'm more of the mindset that if it works, why add more complexity?

Priyashetty17: Israel, I agree with Dan. I've tested this and it works fine. So, why can't we just implement it as it is in this PR?

igomez06: It boils down to it being a member function in Wrangler Context rather than it being it in two places in rancher/client does that make sense? It same thing I'm just saying to move it to Wrangler Context instead of here. It's not adding complexity just organizing it better. 

It makes no sense have client.WranglerContext and client.GetDownstream() this is about design. 

If you notice we don't do that with the dynamic client either.

Priyashetty17: Israel, Are you asking the function be simply moved to : https://github.com/rancher/shepherd/blob/main/pkg/wrangler/context.go?

igomez06: yes, so it's member function of the Wrangler context. Because the you implemented we are treating the downstream wrangler context and the local cluster wrangler context differently. Why would we do that?

Priyashetty17: I've moved it to the wrangler context. Could you please re-review? If there are further adjustments needed to align it with how ProxyDownstream is implemented, it would be helpful if the PIT crew could assist, as I'm not very familiar with it.

pkg/wrangler/context.go
```diff
@@ -246,3 +247,16 @@ func syncOnlyChangedObjects(option controllerContextType) bool {
 	}
 	return false
 }
+
+// GetClusterContext creates a wrangler context to interact with a specific cluster.
+func (w *Context) GetClusterContext(clusterID string) (*Context, error) {
```
igomez06, line 252: This looks good Priya, my only feedback is call it `DownStreamClusterContext`, without the get.

Priyashetty17: Thanks for the review Israel :) I updated the function name to `DownStreamClusterWranglerContext`

caliskanugur (approved):
Looks good to me

floatingman (approved):

igomez06 (approved):
LGTM could you back port this to v2.8 as well?

Priyashetty17:
@igomez06 Here's the backport PR for 2.8: https://github.com/rancher/shepherd/pull/180

PR #167, opened 2024-04-26
[2.8] Update Shepherd Go Mod Packages for 1.29

caliskanugur:
- Bump dependencies to align with 1.29 PR
- Fixes VolumeResourceRequirements change
- Updates pkg/apis and uses 1.29 PR's pkg/apis. 
- Code gens after third step

PR #179, opened 2024-05-07
[Main] Update Shepherd Go Mod Packages for 1.29

caliskanugur:
Issue: https://github.com/rancher/qa-tasks/issues/1304

- Bump dependencies to align with 1.29 PR
- Fixes VolumeResourceRequirements change
- Updates pkg/apis and uses 1.29 PR's pkg/apis. 
- Code gens after third step

Additionally: 

- Fixes a bug that happens in codegen because of the paths
- Adds pkg/generated controller full dir deletion to the existing codegen cleanup

TODO:

- [x] Replaced Rancher dependencies to get updated after they are merged to various Rancher repositories
- [x] After pkg/apis bumps also happen, codegen needs to happen again as this one is pointed to PR's pkg/apis rather than Rancher's pkg/apis


caliskanugur:
Backport: https://github.com/rancher/shepherd/pull/167

markusewalker (approved):
LGTM

pkg/codegen/generator/cleanup/main.go
```diff
@@ -11,8 +11,13 @@ func main() {
 }
 
 func run() error {
-	if err := os.RemoveAll("../../clients/rancher/generated"); err != nil {
+	if err := os.RemoveAll("./pkg/generated"); err != nil {
 		return err
 	}
+
+	if err := os.RemoveAll("./clients/rancher/generated"); err != nil {
```
anupama2501, line 18: shouldn't this be `("../../clients/rancher/generated")` ?

pkg/schemas/management.cattle.io/v3/schema.go
```diff
@@ -326,11 +327,11 @@ func authzTypes(schemas *types.Schemas) *types.Schemas {
 		}).
 		MustImport(&Version, v3.GlobalRoleBinding{}).
 		MustImport(&Version, v3.RoleTemplate{}).
-		MustImport(&Version, v3.PodSecurityPolicyTemplate{}).
-		MustImportAndCustomize(&Version, v3.PodSecurityPolicyTemplateProjectBinding{}, func(schema *types.Schema) {
-			schema.CollectionMethods = []string{http.MethodGet, http.MethodPost}
-			schema.ResourceMethods = []string{}
-		}).
+		// MustImport(&Version, v3.PodSecurityPolicyTemplate{}).
```
anupama2501, line 330: qq why are we commenting these here and in lines 305 - 307?

pkg/codegen/generator/cleanup/main.go
caliskanugur: Great question, the previous version wasn't cleaning up properly, and the path is not relative because of this: https://github.com/rancher/shepherd/blob/main/generator.go#L2

pkg/schemas/management.cattle.io/v3/schema.go
caliskanugur: I think it's a great point, however, the v3 management client needs to be updated - there's another ticket for it, as it's not in the scope of this PR, I think deleting the commented PSP templates would be safer after that, what do you think?

anupama2501: sure, we can remove the code later in that ticket.

anupama2501 (approved):

floatingman (approved):

PR #194, opened 2024-05-20
[v2.9] Fix RKE1 nodetemplates to include cloud credentials

markusewalker:
### Description
In the Rancher UI, the expected workflow for RKE1 nodetemplates is that a user uses an existing cloud credential ID. Our automation does not reflect this and bypasses creating a cloud credential. This is not realistic to what customers do, so this small PR fixes that.

### DIFF
--- a/extensions/rke1/nodetemplates/aws/create.go
+++ b/extensions/rke1/nodetemplates/aws/create.go
@@ -3,6 +3,7 @@ package nodetemplates
 import (
 	"github.com/rancher/shepherd/clients/rancher"
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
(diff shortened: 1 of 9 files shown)


markusewalker:
> should we remove the token stuff from all the individual node provider node template configs?

@igomez06 looking at the configs now, not sure I'm seeing what you mean? I don't mind removing it, just not sure where it is.

markusewalker:
Ahh I see, got it. Yeah, let me go and remove those and push a commit.

slickwarren (approved):

slickwarren (approved):

Josh-Diamond (approved):

Josh-Diamond (approved):

igomez06 (approved):
LGTM

PR #197, opened 2024-05-21
RKE2 template test

susesgartner:
Release testing PR for rke2 templates.

Issue: https://github.com/rancher/qa-tasks/issues/542

### DIFF
--- a/extensions/charts/rke2template.go
+++ b/extensions/charts/rke2template.go
@@ -1,0 +1,129 @@
+package charts
+
+import (
+	"context"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/extensions/cloudcredentials"
+	"github.com/rancher/shepherd/extensions/clusters"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/projects"
+	"github.com/rancher/shepherd/extensions/wait"
+	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+	"github.com/rancher/shepherd/pkg/namegenerator"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	fleetNamespace = "fleet-default"
+	localCluster   = "local"
+)
+
+// InstallTemplateChart installs a template from a repo.
+func InstallTemplateChart(client *rancher.Client, repoName, templateName, clusterName, k8sVersion string, credentials *cloudcredentials.CloudCredential) error {
+	latestVersion, err := client.Catalog.GetLatestChartVersion(templateName, repoName)
+	if err != nil {
+		return err
+	}
+
+	project, err := projects.GetProjectByName(client, localCluster, "System")
+	if err != nil {
+		return err
+	}
+
+	installOptions := &InstallOptions{
+		Cluster: &clusters.ClusterMeta{
+			ID: localCluster,
+		},
+		Version:   latestVersion,
+		ProjectID: project.ID,
+	}
+
+	serverSetting, err := client.Management.Setting.ByID(serverURLSettingID)
+	if err != nil {
+		return err
+	}
+
+	registrySetting, err := client.Management.Setting.ByID(defaultRegistrySettingID)
+	if err != nil {
+		return err
+	}
+
+	chartInstallActionPayload := &payloadOpts{
+		InstallOptions:  *installOptions,
+		Name:            templateName,
+		Namespace:       fleetNamespace,
+		Host:            serverSetting.Value,
+		DefaultRegistry: registrySetting.Value,
+	}
+
+	chartValues, err := client.Catalog.GetChartValues(repoName, templateName, installOptions.Version)
+	if err != nil {
+		return err
+	}
+
+	chartInstallAction := TemplateInstallAction(chartInstallActionPayload, repoName, clusterName, credentials.ID, k8sVersion, fleetNamespace, chartValues)
+
+	catalogClient, err := client.GetClusterCatalogClient(installOptions.Cluster.ID)
+	if err != nil {
+		return err
+	}
+
+	err = client.Catalog.InstallChart(chartInstallAction, repoName)
+	if err != nil {
+		return err
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
+		err := client.Catalog.UninstallChart(templateName, fleetNamespace, newChartUninstallAction())
+		if err != nil {
+			return err
+		}
+
+		watchAppInterface, err := catalogClient.Apps(fleetNamespace).Watch(context.TODO(), metav1.ListOptions{
+			FieldSelector:  "metadata.name=" + templateName,
+			TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+		})
+		if err != nil {
+			return err
+		}
+
+		err = wait.ResourceDelete(watchAppInterface)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	err = VerifyChartInstall(catalogClient, fleetNamespace, templateName)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+// TemplateInstallAction creates the payload used when installing a template chart
+func TemplateInstallAction(InstallActionPayload *payloadOpts, repoName, clusterName, cloudCredential, k8sVersion, namespace string, chartValues map[string]any) *types.ChartInstallAction {
+	chartValues["cloudCredentialSecretName"] = cloudCredential
+	chartValues["kubernetesVersion"] = k8sVersion
+	chartValues["cluster"].(map[string]any)["name"] = clusterName
+
+	for index := range chartValues["nodepools"].(map[string]any) {
+		chartValues["nodepools"].(map[string]any)[index].(map[string]any)["name"] = namegenerator.AppendRandomString("nodepool")
+	}
+
+	chartInstall := newChartInstall(
+		InstallActionPayload.Name,
+		InstallActionPayload.InstallOptions.Version,
+		InstallActionPayload.InstallOptions.Cluster.ID,
+		InstallActionPayload.InstallOptions.Cluster.Name,
+		InstallActionPayload.Host,
+		repoName,
+		InstallActionPayload.InstallOptions.ProjectID,
+		InstallActionPayload.DefaultRegistry,
+		chartValues)
+	chartInstalls := []types.ChartInstall{*chartInstall}
+
+	return newChartInstallAction(namespace, InstallActionPayload.ProjectID, chartInstalls)
+}
--- a/extensions/provisioning/verify.go
+++ b/extensions/provisioning/verify.go
@@ -165,8 +165,10 @@ func VerifyCluster(t *testing.T, client *rancher.Client, clustersConfig *cluster
 	podErrors := pods.StatusPods(client, status.ClusterName)
 	assert.Empty(t, podErrors)
 
-	if clustersConfig.ClusterSSHTests != nil {
-		VerifySSHTests(t, client, cluster, clustersConfig.ClusterSSHTests, status.ClusterName)
+	if clustersConfig != nil {
+		if clustersConfig.ClusterSSHTests != nil {
+			VerifySSHTests(t, client, cluster, clustersConfig.ClusterSSHTests, status.ClusterName)
+		}
 	}
 }
 
--- a/extensions/steve/steve.go
+++ b/extensions/steve/steve.go
@@ -6,12 +6,14 @@ import (
 	"time"
 
 	"github.com/rancher/shepherd/clients/rancher"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/sirupsen/logrus"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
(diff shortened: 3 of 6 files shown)


Josh-Diamond:
LGTM Sam! great work!

Two questions (nits?):
- on line 42 of `rke2template.go`, the following line of code is seen: `k8sversions, err := kubernetesversions.Default(client, "rke2", nil)`, do we want to store the string "rke2" as a const? -- its unclear to me what our convention around this is, or if we have a procedure in place around what is acceptable for params, so just pointing that out 
- second point is also in `rke2teamplate.go`, I noticed "fmt" is being imported instead of logrus.  Is this intentional? doesn't cause alarm on my end, and i vaguely remember issues myself where logrus.Error was having a linter meltdown when i tried returning that in previous qa-tasks, but I also recall we decided sometime back to use `logrus` whenever possible. So just wanted to point that out as well, and see if there was a specific reason "fmt" was being used 🙂

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/rancher/shepherd/pkg/wait"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
+	fleetNamespace = "fleet-default"
+)
+
+// InstallAWSOutOfTreeChart installs the CSI chart for aws cloud provider in a given cluster.
```
sowmyav27, line 22: I don't think this comment in appropriate here. Can we modify this to reflect for the func `InstallRKE2TemplateChart` ?

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/rancher/shepherd/extensions/clusters/kubernetesversions"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+	"github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/shepherd/pkg/wait"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
```
sowmyav27, line 18: Can you explain why are we installing in `fleetNamespace` ?

extensions/charts/rke2template.go
markusewalker, line 22: nit - probably just a copy and paste, but this needs to be reflective of `InstallRKE2TemplateChart` and not the AWS out of tree chart.

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+
+	chartInstallActionPayload := &payloadOpts{
+		InstallOptions:  *installOptions,
+		Name:            templateName,
+		Namespace:       fleetNamespace,
+		Host:            serverSetting.Value,
+		DefaultRegistry: registrySetting.Value,
+	}
+
+	k8sversions, err := kubernetesversions.Default(client, "rke2", nil)
```
markusewalker, line 42: nit - can you put `rke2` in the const block above?

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+	})
+
+	err = VerifyChartInstall(catalogClient, fleetNamespace, templateName)
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func RKE2TemplateInstallAction(InstallActionPayload *payloadOpts, repoName, chartName, cloudCredential, k8sVersion, namespace string, chartValues map[string]any) *types.ChartInstallAction {
```
markusewalker, line 100: Can you add a function description here for godocs?

extensions/provisioning/verify.go
```diff
@@ -165,8 +165,10 @@ func VerifyCluster(t *testing.T, client *rancher.Client, clustersConfig *cluster
 	podErrors := pods.StatusPods(client, status.ClusterName)
 	assert.Empty(t, podErrors)
 
-	if clustersConfig.ClusterSSHTests != nil {
-		VerifySSHTests(t, client, cluster, clustersConfig.ClusterSSHTests, status.ClusterName)
+	if clustersConfig != nil {
```
markusewalker, line 168: Is this parent conditional necessary? What was the issue with just having this left as is?

markusewalker (commented):
Left some comments. Overall, LGTM.

sowmyav27 (changes requested):

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
+		err := client.Catalog.UninstallChart(templateName, fleetNamespace, newChartUninstallAction())
+		if err != nil {
+			return err
+		}
+
+		watchAppInterface, err := catalogClient.Apps(fleetNamespace).Watch(context.TODO(), metav1.ListOptions{
```
sowmyav27, line 70: Is this piece of code for `Verify UninstallChart`? If yes, it should go in a separate function.

sowmyav27 (changes requested):

extensions/charts/rke2template.go
markusewalker: I actually think this is fine. This is part of the `RegisterCleanupFunc`, so if your config has `cleanup: true`, then it's going to uninstall the chart. We do this throughout the framework, so I don't think we should change it here

extensions/charts/rke2template.go
susesgartner: I'll update that.

extensions/provisioning/verify.go
susesgartner: Since rke2 templates calls verify cluster without a clusterconfig the test would fail at this point since clusterConfig.ClusterClusterSSHTests was attempting to reference a nil object.

extensions/charts/rke2template.go
susesgartner: I will fix this.

extensions/charts/rke2template.go
susesgartner: This is the behavior that rke2 templates uses when installing through the UI.

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func RKE2TemplateInstallAction(InstallActionPayload *payloadOpts, repoName, chartName, cloudCredential, k8sVersion, namespace string, chartValues map[string]any) *types.ChartInstallAction {
+	chartValues["cloudCredentialSecretName"] = cloudCredential
+	chartValues["kubernetesVersion"] = k8sVersion
+	chartValues["cluster"].(map[string]any)["name"] = InstallActionPayload.Name
```
slickwarren, line 103: isn't there an object you can use / cast to instead of using raw `map[string]any`?

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+
+	chartValues, err := client.Catalog.GetChartValues(repoName, templateName, installOptions.Version)
+	if err != nil {
+		return err
+	}
+
+	chartInstallAction := RKE2TemplateInstallAction(chartInstallActionPayload, repoName, templateName, credentials.ID, k8sversions[0], fleetNamespace, chartValues)
```
slickwarren, line 52: nit; this can be addressed when we go back and add the upgrade use case if you want but, it might be better to pass in the k8s version as a variable to the function instead of using the latest version available and getting that in this function as it is now.

extensions/charts/rke2template.go
igomez06: This part of the clean up, so this looks good.

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,122 @@
...
+			FieldSelector:  "metadata.name=" + templateName,
+			TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+		})
+		if err != nil {
+			return err
+		}
+
+		err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+			if event.Type == watch.Error {
+				return false, fmt.Errorf("there was an error uninstalling rke2 template chart")
```
igomez06, line 80: Sam since this is an Errorf, can you add the error return as part of the message?

extensions/charts/rke2template.go
susesgartner: Not that I am aware of.

extensions/wait/wait.go
```diff
@@ -0,0 +1,33 @@
+package wait
+
+import (
+	"github.com/rancher/shepherd/pkg/wait"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+func ResourceCreation(watchInterface watch.Interface) error {
```
slickwarren, line 8: nit; the name I would expect is `ResourceCreate` and `ResourceDelete` though I don't know if we have rules around this. Most other names follow this, though.

extensions/charts/rke2template.go
```diff
@@ -0,0 +1,139 @@
...
+
+		watchAppInterface, err := catalogClient.Apps(fleetNamespace).Watch(context.TODO(), metav1.ListOptions{
+			FieldSelector:  "metadata.name=" + templateName,
+			TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+		})
+		if err != nil {
+			return err
+		}
+
+		err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
```
slickwarren, line 94: couldn't this use `wait.ResourceDelete(ion)`?

slickwarren (commented):
overall lgtm, some smaller questions.

extensions/wait/wait.go
susesgartner: Good point. I'll rename them

extensions/charts/rke2template.go
susesgartner: Good point.

extensions/wait/wait.go
```diff
@@ -0,0 +1,33 @@
+package wait
+
+import (
+	"github.com/rancher/shepherd/pkg/wait"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+func ResourceCreate(watchInterface watch.Interface) error {
```
igomez06, line 8: Can we get go doc comments here please

extensions/charts/rke2template.go
sowmyav27: I understand it is part of clean up. I don't understand why it cannot be another function ?

sowmyav27 (changes requested):

extensions/charts/rke2template.go
markusewalker: I actually agree with you that it probably should be a separate function that all of the charts can reference. IMO though, probably better if that's a separate PR just to not over bloat this PR

extensions/wait/wait.go
susesgartner: Added some comments as per your suggestion.

extensions/charts/rke2template.go
susesgartner: I also agree that moving it to another function may be a good idea. But if i'm going to create such a function I should probably update the rest of charts for consistency and that should be handled in a standalone PR.

markusewalker (approved):
LGTM

go.mod
```diff
@@ -41,7 +41,7 @@ require (
 	golang.org/x/crypto v0.23.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.29.4
-	k8s.io/apimachinery v0.29.4
+	k8s.io/apimachinery v0.30.1
```
igomez06, line 44: Quick this does match release/v2.9 version?

extensions/steve/steve.go
```diff
@@ -36,3 +38,45 @@ func WaitForSteveResourceDeletion(client *rancher.Client, interval, timeout time
...
+		}
+
+		return false, nil
+	})
+
+	return err
+}
+
+// CreateResource creates a steve resource and polls the resulting object until it comes up active.
+func CreateResource(client *rancher.Client, v1ResourceType string, v1Resource any, poll bool, interval, timeout time.Duration) (*v1.SteveAPIObject, error) {
```
igomez06, line 67: Can we name this `CreateAndWaitForResource`? Sort of to distinguish why some one would choose this extension over the client create. In the near future we will have generated steve clients as well, just an FYI

go.mod
susesgartner: No it doesn't look like they match.

igomez06: We don't need this version bump then right?

susesgartner: I don't believe so but i'll look into it.

extensions/steve/steve.go
susesgartner: I am fine with this, I'll update it.

go.mod
susesgartner: Fixed this to not include the version bump.

igomez06 (approved):
LGTM

igomez06:
@susesgartner this requires back ports right?

sowmyav27 (approved):
lgtm

slickwarren (approved):

PR #198, opened 2024-05-22
[v2.9] Enhance existing hardened clusters tests to install CIS chart

markusewalker:
### PR Description
As of K8s 1.25, PSPs were deprecated. As such, PSAs came to be the official security that K8s clusters followed. Rancher no longer supports K8s 1.24, so no need to have PSPs in our automation anymore.

Additionally, part of the workflow to ensure a cluster has been hardened is to install the CIS Benchmark chart and run a scan. We do not do that, so the workflow is not 100% indicative that the downstream cluster is indeed hardened.


extensions/charts/cisbenchmark.go
```diff
@@ -0,0 +1,298 @@
...
+	chartInstallCRD := newChartInstall(p.Name+"-crd", p.Version, p.Cluster.ID, p.Cluster.Name, p.Host, rancherChartsName, p.ProjectID, p.DefaultRegistry, nil)
+	chartInstalls := []types.ChartInstall{*chartInstallCRD, *chartInstall}
+
+	chartInstallAction := newChartInstallAction(p.Namespace, p.ProjectID, chartInstalls)
+
+	return chartInstallAction, nil
+}
+
+// UpgradeCISBenchmarkChart is a helper function that upgrades the cis-benchmark chart.
+func UpgradeCISBenchmarkChart(client *rancher.Client, installOptions *InstallOptions) error {
```
slickwarren, line 201: do you run upgrade tests anywhere yet? or is this just for support later?

slickwarren (commented):
one question, but overall lgtm. Awaiting rke2 run for approval

extensions/charts/cisbenchmark.go
markusewalker: This is support for later, correct.

extensions/hardening/k3s/harden_nodes.go
```diff
@@ -40,14 +37,12 @@ func HardenNodes(nodes []*nodes.Node, nodeRoles []string, kubeVersion string) er
...
 		if strings.Contains(nodeRoles[key], "--controlplane") {
-			logrus.Infof("Copying over files to node %s", node.NodeID)
-			user, err := user.Current()
-			if err != nil {
-				return nil
-			}
-
-			dirPath := filepath.Join(user.HomeDir, "go/src/github.com/rancher/shepherd/extensions/hardening/k3s")
-			err = node.SCPFileToNode(dirPath+"/audit.yaml", "/home/"+node.SSHUser+"/audit.yaml")
+			_, err = node.ExecuteCommand(`sudo bash -c 'cat << EOF > /home/` + node.SSHUser + `/audit.yaml
```
slickwarren, line 40: I think this yaml should live in an actual file that's then read into this call instead of raw string.

extensions/clusters/clusters.go
```diff
@@ -868,32 +856,31 @@ func HardenK3SClusterConfig(clusterName, namespace string, clustersConfig *Clust
...
-				Config: rkev1.GenericMap{
-					Data: map[string]interface{}{
-						"profile":                 "cis-1.23",
-						"protect-kernel-defaults": true,
-					},
+	v1Cluster.Spec.RKEConfig.MachineSelectorConfig = []rkev1.RKESystemConfig{
+		{
+			Config: rkev1.GenericMap{
+				Data: map[string]interface{}{
+					"profile":             "cis",
```
slickwarren, line 879: is this going to change frequently? Just noticed that this is `cis`, it used to be `cis-1.6` or `cis-1.23`, and the test name is `cis 1.8`. Should this be read in from the scan name or something instead of hard coding it?

extensions/clusters/clusters.go
```diff
@@ -868,32 +856,31 @@ func HardenK3SClusterConfig(clusterName, namespace string, clustersConfig *Clust
 	return v1Cluster
 }
 
-// HardenRKE2ClusterConfig is a constructor for a apisV1.Cluster object, to be used by the rancher.Client.Provisioning client.
+// HardenRKE1ClusterConfig is a function that modifies the cluster configuration to be hardened according to the CIS benchmark.
+func HardenRKE1ClusterConfig(client *rancher.Client, clusterName string, clustersConfig *ClusterConfig) *management.Cluster {
+	cluster := NewRKE1ClusterConfig(clusterName, client, clustersConfig)
+
+	cluster.DefaultPodSecurityAdmissionConfigurationTemplateName = rancherRestricted
+	cluster.RancherKubernetesEngineConfig.Services.Etcd.GID = 52034
```
slickwarren, line 864: seeing these as raw values is strange to me. Can you make it a const and maybe even add a comment as to why it is what it is, or a ref in the rancher/rancher code to where this is set? might be overkill but if we ever have to change it, this info would be nice to have.

extensions/clusters/clusters.go
```diff
@@ -824,32 +825,19 @@ func AgentAffinityConfigHelper(advancedClusterAffinity *management.Affinity) *co
...
-			"audit-log-path=/var/lib/rancher/k3s/server/logs/audit.log",
-			"audit-log-maxage=30",
-			"audit-log-maxbackup=10",
-			"audit-log-maxsize=100",
-			"request-timeout=300s",
-			"service-account-lookup=true",
-		}
-	} else {
-		v1Cluster.Spec.RKEConfig.MachineGlobalConfig.Data["kube-apiserver-arg"] = []string{
-			"admission-control-config-file=/var/lib/rancher/k3s/server/psa.yaml",
```
slickwarren, line 844: was this not supposed to be here before, or did we miss moving it over? (why was this rm'd, since it used to be in the 'else' block)

extensions/charts/cisbenchmark.go
slickwarren, line 201: why does this not need a cleanup func?

extensions/hardening/k3s/harden_nodes.go
markusewalker: Jenkins specifically was calling this an issue and couldn't find the file. Removing it was the only way it passed successfully. Locally, it works. through Jenkins, it doesn't.

extensions/clusters/clusters.go
markusewalker: `cis` specifically was introduced to address this concern. Moving forward, `cis` is what is documented that users should be using so that with future profiles, this won't change.

Upgrading as an example is a huge plus as to keeping this.

extensions/clusters/clusters.go
markusewalker: I can put it into a const. I personally think it's too much to comment it. The hardening guides we put out to customers go into extreme detail as to all of this, so I would expect someone to read that.

extensions/clusters/clusters.go
markusewalker: This is being removed because Rancher sets this when you set the PSACT. It's redundant having it here.

extensions/charts/cisbenchmark.go
markusewalker: I was just following the practices from the other charts for the upgrade charts. I am not sure why they are like that, but I had it this way for consistency.

extensions/hardening/rke1/harden_nodes.go
```diff
@@ -0,0 +1,90 @@
...
+// HardenRKE1Nodes hardens the nodes by setting kernel parameters and creating the etcd user and group
+func HardenRKE1Nodes(nodes []*nodes.Node, nodeRoles []string) error {
+	for _, node := range nodes {
+		logrus.Infof("Setting kernel parameters on node: %s", node.NodeID)
+		_, err := node.ExecuteCommand("sudo bash -c 'echo vm.panic_on_oom=0 >> /etc/sysctl.d/90-kubelet.conf'")
+		if err != nil {
+			return err
+		}
+
+		_, err = node.ExecuteCommand("sudo bash -c 'echo vm.overcommit_memory=1 >> /etc/sysctl.d/90-kubelet.conf'")
```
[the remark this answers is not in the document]
markusewalker: I am not sure of the documented OS, I would say no? I see where you're going with that, I personally don't think that's enough of a concern to not have it here. Our automation typically runs through Ubuntu, RHEL or SLES and each of those don't fit that issue IMO.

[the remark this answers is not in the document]
markusewalker: RKE1/RKE2 share the file names, that's true, but they're in different packages. I ended up putting separate consts in their respective files. I thought it would look cleaner than having to define elsewhere.

extensions/charts/cisbenchmark.go
```diff
@@ -0,0 +1,298 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// newCISBenchmarkChartInstallAction is a private helper function that returns chart install action with CIS benchmark and payload options.
+func newCISBenchmarkChartInstallAction(p *payloadOpts) (*types.ChartInstallAction, error) {
```
caliskanugur, line 190: I've got two questions here:

1. Does CIS Benchmark have any options - like Jaeger with Istio? (if yes) is it expected to have a struct for future tests, if not ignore this one, please.
2. Does this default template `ChartInstall` payload value provide support on the K3s, RKE1, and RKE2 types? So installation on any provider type is out-of-the-box?

extensions/charts/cisbenchmark.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+	"github.com/rancher/shepherd/pkg/wait"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
+	CISBenchmarkNamespace = "cis-operator-system"
+	CISBenchmarkName      = "rancher-cis-benchmark"
+	CISBenchmarkCRDName   = "rancher-cis-benchmark-crd"
```
caliskanugur, line 22: Nit; these need comments, examples should be in other charts files. These are going to be exposed with Godoc.

caliskanugur (commented):
The charts part looks good to me, I've got one question and a nit

extensions/charts/cisbenchmark.go
markusewalker: 1. It's not expected to have a struct, correct.
2. That's correct, cluster type is agnostic here. `ChartInstall` is provided for any provider type out of the box, which is makes it pretty straightforward.

extensions/charts/cisbenchmark.go
markusewalker: Sure, I'll add them.

markusewalker: Resolved

extensions/hardening/k3s/harden_nodes.go
markusewalker: Talked offline on this point. This is most likely being moved to actions once that becomes a thing. When that happens, I'll convert this back into files. Jenkins won't have this issue once that happens.

slickwarren (approved):

caliskanugur (approved):
Looks good to me

PR #200, opened 2024-05-23
[v2.8] Modify hosted provider create methods to not read directly from config file

valaparthvi:
This PR modifies the hosted provider create functions to not read directly from the CATTLE_TEST_CONFIG file but obtain the data instead as a user input. This makes the functions more generic and useful when the user needs to run parallel test and the config file needs to be modified during the test, these changes will help in avoiding race condition with the resource.

Note: This is a breaking change PR.

### DIFF
--- a/extensions/clusters/aks/aks_cluster_config.go
+++ b/extensions/clusters/aks/aks_cluster_config.go
@@ -2,7 +2,6 @@ package aks
 
 import (
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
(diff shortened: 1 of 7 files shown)


valaparthvi:
Marked the test using these methods as deprecated/skip on rancher/rancher.
https://github.com/rancher/rancher/pull/45584

valaparthvi:
@igomez06 Can I get an approval on this one?

igomez06 (approved):
LGTM

PR #211, opened 2024-06-12
adding support to upgrade out-of-tree charts for rke1 vsphere

slickwarren:
supports: https://github.com/rancher/rancher/pull/45792

updated the way that pulling data from the vsphereConfig works, and added some functionality around upgrading + validating charts.

### DIFF
--- a/extensions/charts/config.go
+++ b/extensions/charts/config.go
@@ -1,0 +1,9 @@
+package charts
+
+const (
+	ConfigurationFileKey = "chartUpgrade"
+)
+
+type Config struct {
+	IsUpgradable bool `json:"isUpgradable" yaml:"isUpgradable"`
+}
--- a/extensions/charts/payloads.go
+++ b/extensions/charts/payloads.go
@@ -80,14 +80,14 @@ func newChartInstall(name, version, clusterID, clusterName, url, repoName, proje
 }
 
 // newChartUpgradeAction is a private constructor that creates a chart upgrade with given chart values that can be used for chart upgrade action.
-func newChartUpgrade(name, version, clusterID, clusterName, url, defaultRegistry string, chartValues map[string]interface{}) *types.ChartUpgrade {
+func newChartUpgrade(chartName, releaseName, version, clusterID, clusterName, url, defaultRegistry string, chartValues map[string]interface{}) *types.ChartUpgrade {
 	chartUpgrade := types.ChartUpgrade{
 		Annotations: map[string]string{
 			"catalog.cattle.io/ui-source-repo":      "rancher-charts",
 			"catalog.cattle.io/ui-source-repo-type": "cluster",
 		},
-		ChartName:   name,
-		ReleaseName: name,
+		ChartName:   chartName,
+		ReleaseName: releaseName,
 		Version:     version,
 		Values: v3.MapStringInterface{
 			"global": map[string]interface{}{
--- a/extensions/charts/verify.go
+++ b/extensions/charts/verify.go
@@ -37,3 +37,57 @@ func VerifyChartInstall(client *catalog.Client, chartNamespace, chartName string
 	})
 	return err
 }
+
+// VerifyChartUpgrade verifies that the app from a chart was successfully upgraded
+func VerifyChartUpgrade(client *catalog.Client, chartNamespace, chartName, chartVersion string) error {
+	watchAppInterface, err := client.Apps(chartNamespace).Watch(context.TODO(), metav1.ListOptions{
+		FieldSelector:  "metadata.name=" + chartName,
+		TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+	})
+	if err != nil {
+		return err
+	}
+
+	err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+		app := event.Object.(*v1.App)
+
+		state := app.Status.Summary.State
+		if state == string(v1.StatusPendingUpgrade) {
+			return true, nil
+		}
+
+		if state == string(v1.StatusFailed) {
+			return false, errors.New("chart upgrade has failed")
+		}
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	watchAppInterface, err = client.Apps(chartNamespace).Watch(context.TODO(), metav1.ListOptions{
+		FieldSelector:  "metadata.name=" + chartName,
+		TimeoutSeconds: &defaults.WatchTimeoutSeconds,
+	})
+	if err != nil {
+		return err
+	}
+
+	err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+		app := event.Object.(*v1.App)
+
+		state := app.Status.Summary.State
+		if state == string(v1.StatusDeployed) {
+			if app.Spec.Chart.Metadata.Version == chartVersion {
+				return true, nil
+			}
+			return false, errors.New("chart not upgraded to expected version: " + chartVersion)
+		}
+
+		if state == string(v1.StatusFailed) {
+			return false, errors.New("chart upgrade has failed")
+		}
+		return false, nil
+	})
+	return err
+}
--- a/extensions/charts/vsphereoutoftree.go
+++ b/extensions/charts/vsphereoutoftree.go
@@ -1,10 +1,16 @@
 package charts
 
 import (
+	"context"
+
+	v1 "github.com/rancher/rancher/pkg/apis/catalog.cattle.io/v1"
+	rancherv1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/clusters"
 	"github.com/rancher/shepherd/extensions/projects"
 	"github.com/rancher/shepherd/extensions/rke1/nodetemplates"
 	"github.com/rancher/shepherd/pkg/api/steve/catalog/types"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/rancher/shepherd/clients/rancher"
 	"github.com/rancher/shepherd/clients/rancher/catalog"
@@ -14,7 +20,9 @@ import (
 const (
 	systemProject       = "System"
 	vsphereCPIchartName = "rancher-vsphere-cpi"
+	manualCPIchartName  = "vsphere-cpi"
 	vsphereCSIchartName = "rancher-vsphere-csi"
+	manualCSIchartName  = "vsphere-csi"
 
 	vcenter      = "vCenter"
 	storageclass = "storageClass"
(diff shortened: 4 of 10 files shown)


extensions/charts/verify.go
```diff
@@ -37,3 +37,66 @@ func VerifyChartInstall(client *catalog.Client, chartNamespace, chartName string
...
+
+	err = wait.WatchWait(watchAppInterface, func(event watch.Event) (ready bool, err error) {
+		app := event.Object.(*v1.App)
+
+		state := app.Status.Summary.State
+		if state == string(v1.StatusDeployed) {
+			if app.Spec.Chart.Metadata.Version == chartVersion {
+				return true, nil
+			}
+			return false, errors.New("chart not upgraded to expected version")
```
markusewalker, line 84: Can you change this to show the expected version it tried to upgrade to in a situation that it fails? This would mainly be for debugging purposes.

extensions/charts/verify.go
```diff
@@ -37,3 +37,66 @@ func VerifyChartInstall(client *catalog.Client, chartNamespace, chartName string
...
+
+		if state == string(v1.StatusFailed) {
+			return false, errors.New("chart upgrade has failed")
+		}
+		return false, nil
+	})
+	return err
+}
+
+func GetExistingChartInstall(client *catalog.Client, clusterName, chartName, chartNamespace string) (*v1.App, error) {
```
markusewalker, line 95: Can you add a godoc for this function?

extensions/charts/vsphereoutoftree.go
```diff
@@ -55,14 +60,30 @@ func InstallVsphereOutOfTreeCharts(client *rancher.Client, vsphereTemplate *node
 		return err
 	}
 
-	latestCPIVersion, err := catalogClient.GetLatestChartVersion(vsphereCPIchartName, catalog.RancherChartRepo)
+	var cpiVersion string
+	var csiVersion string
+	cpiVersions, err := catalogClient.GetListChartVersions(vsphereCPIchartName, catalog.RancherChartRepo)
```
markusewalker, line 65: nit - add a new line for code readability.

extensions/charts/vsphereoutoftree.go
```diff
@@ -73,31 +94,29 @@ func InstallVsphereOutOfTreeCharts(client *rancher.Client, vsphereTemplate *node
 		Host:            serverSetting.Value,
 		DefaultRegistry: registrySetting.Value,
 	}
+	vsphereTemplateConfig := r1vsphere.GetVsphereNodeTemplate()
```
markusewalker, line 97: nit - please add an extra line for code readability.

extensions/charts/vsphereoutoftree.go
```diff
@@ -124,17 +144,17 @@ func InstallVsphereOutOfTreeCharts(client *rancher.Client, vsphereTemplate *node
 }
 
 // vsphereCPIChartInstallAction is a helper function that returns a chartInstallAction for aws out-of-tree chart.
```
markusewalker, line 146: This comment looks like it was copy paste since it says `aws out-of-tree`. This needs to be updated for vsphere.

extensions/charts/vsphereoutoftree.go
```diff
@@ -152,20 +172,20 @@ func vsphereCPIChartInstallAction(client *catalog.Client, chartInstallActionPayl
 }
 
 // vsphereCSIChartInstallAction is a helper function that returns a chartInstallAction for aws out-of-tree chart.
```
markusewalker, line 174: Same comment as above.

extensions/charts/vsphereoutoftree.go
```diff
@@ -181,3 +201,216 @@ func vsphereCSIChartInstallAction(client *catalog.Client, chartInstallActionPayl
...
+		return err
+	}
+
+	catalogClient, err := client.GetClusterCatalogClient(cluster.ID)
+	if err != nil {
+		return err
+	}
+
+	cpiChartName := vsphereCPIchartName
+	// Change this name to set var. to manual if fails
```
markusewalker, line 233: Is this meant to be here? Or was this just for troubleshooting purposes?

extensions/charts/vsphereoutoftree.go
```diff
@@ -181,3 +201,216 @@ func vsphereCSIChartInstallAction(client *catalog.Client, chartInstallActionPayl
...
+
+	csiChartName := vsphereCSIchartName
+	csiVersion, err := catalogClient.GetLatestChartVersion(csiChartName, catalog.RancherChartRepo)
+	if err != nil {
+		return err
+	}
+
+	cpiChartObject, err := GetExistingChartInstall(catalogClient, clusterName, cpiChartName, kubeSystemNamespace)
+	if err != nil {
+
```
markusewalker, line 247: nit - remove this extra line of space.

extensions/charts/vsphereoutoftree.go
```diff
@@ -181,3 +201,216 @@ func vsphereCSIChartInstallAction(client *catalog.Client, chartInstallActionPayl
...
+	cpiSpec := new(v1.ReleaseSpec)
+	err = rancherv1.ConvertToK8sType(cpiChartObject.Spec, &cpiSpec)
+	if err != nil {
+		return err
+	}
+
+	csiChartObject, err := GetExistingChartInstall(catalogClient, clusterName, csiChartName, kubeSystemNamespace)
+	if err != nil {
+
```
markusewalker, line 263: Same comment as above with the excess whitespace.

extensions/charts/vsphereoutoftree.go
```diff
@@ -181,3 +201,216 @@ func vsphereCSIChartInstallAction(client *catalog.Client, chartInstallActionPayl
...
+		logrus.Infof(
+			"No chart version to upgrade to, already on the latest CSI version: %s vs installed %s",
+			csiVersion,
+			csiSpec.Chart.Metadata.Version)
+	}
+
+	return err
+}
+
+// vsphereCPIChartUpgradeAction is a helper function that returns a chartUpgradeAction for aws out-of-tree chart.
```
markusewalker, line 361: This needs to be updated to say vsphere out of tree, not aws out of tree.

extensions/charts/vsphereoutoftree.go
```diff
@@ -181,3 +201,216 @@ func vsphereCSIChartInstallAction(client *catalog.Client, chartInstallActionPayl
...
+		chartUpgradeActionPayload.Cluster.Name,
+		repoName,
+		chartUpgradeActionPayload.DefaultRegistry,
+		chartValues)
+	chartUpgrades := []types.ChartUpgrade{*chartUpgrade}
+
+	return newChartUpgradeAction(chartNamespace, chartUpgrades), nil
+}
+
+// vsphereCSIChartUpgradeAction is a helper function that returns a chartUpgradeAction for aws out-of-tree chart.
```
markusewalker, line 388: Same comment as above.

markusewalker (commented):
Left some comments and questions.

extensions/charts/config.go
```diff
@@ -0,0 +1,9 @@
+package charts
+
+const (
+	ConfigurationFileKey = "chartUpgrade"
+)
+
+type Config struct {
+	IsUpgradable bool `json:"isUpgradable" yaml:"isUpgradable"`
```
igomez06, line 8: I need more explanation on why this is added now. It's not quite clear.

slickwarren: By default when installing a chart, we install the latest version. We need a way to know when the user wants to install a version of the chart that can be upgraded i.e. by installing latest-1 version so that the chart can be upgraded later.

markusewalker (approved):

extensions/charts/config.go
caliskanugur: I've got the same confusion, we have upgrade test cases for other chart tests, why do we need to run the same test with a different test case,  is there any difference with the existing chart test cases?

extensions/charts/payloads.go
```diff
@@ -80,14 +80,14 @@ func newChartInstall(name, version, clusterID, clusterName, url, repoName, proje
...
+func newChartUpgrade(chartName, releaseName, version, clusterID, clusterName, url, defaultRegistry string, chartValues map[string]interface{}) *types.ChartUpgrade {
 	chartUpgrade := types.ChartUpgrade{
 		Annotations: map[string]string{
 			"catalog.cattle.io/ui-source-repo":      "rancher-charts",
 			"catalog.cattle.io/ui-source-repo-type": "cluster",
 		},
-		ChartName:   name,
-		ReleaseName: name,
+		ChartName:   chartName,
+		ReleaseName: releaseName,
```
caliskanugur, line 90: It seems like the added argument is still the same correct? Does this affect anything on the test end or just another way of adding the same parameter?

extensions/charts/verify.go
```diff
@@ -37,3 +37,67 @@ func VerifyChartInstall(client *catalog.Client, chartNamespace, chartName string
...
+			return false, errors.New("chart upgrade has failed")
+		}
+		return false, nil
+	})
+	return err
+}
+
+// GetExistingChartInstall returns the App object of an installed chart.
+func GetExistingChartInstall(client *catalog.Client, clusterName, chartName, chartNamespace string) (*v1.App, error) {
+	appObject, err := client.Apps(chartNamespace).Get(context.TODO(), chartName, metav1.GetOptions{})
```
caliskanugur, line 97: I am unsure why we need this for a single client wrap, what's the pros/cons of adding this function?

caliskanugur (commented):
A few questions

extensions/charts/verify.go
igomez06: I definitely agree on this. Is this really needed. The client call is already a single liner, this just seems like more bloat.

extensions/charts/config.go
slickwarren: The big difference here is cloud provider is fundamentally closer to kubernetes_version than it is to a normal chart install / upgrade. So for provisioning, the upgrade packages depends on a cluster being provided. Therefore the cloud provider depends on a cluster being provided that has charts ready for upgrade. This introduced the reason for the new variable to be used during provisioning, because we need a way to install an upgradable (latest - 1 version) chart during provisioning so that we can run an upgrade test later. 

I met with Israel and he understood this, but if its still unclear / pushback I can sync with you as well when you have time.

extensions/charts/payloads.go
slickwarren: this was a difference I found if the chart is installed manually vs. via automation. Automation seems to have chartName == releaseName. But manually deploying some charts like vsphere CPI / CSI sets the releaseName = chartName - `rancher-`.  which was causing problems with the upgrade. Since this involves a provisioning test, I wanted to update this so that the user could provide a cluster that was provisioned manually and/or via automation. This was the only way I could see to do that dynamically.

extensions/charts/verify.go
slickwarren: sure, updated / rm'd from public funciton list

igomez06 (approved):
LGTM

PR #224, opened 2024-06-21
vai config

brudnak:
Able to specify vai status in the config

### DIFF
--- a/extensions/vai/config.go
+++ b/extensions/vai/config.go
@@ -1,0 +1,9 @@
+package vai
+
+const (
+	ConfigurationFileKey = "vai"
+)
+
+type Config struct {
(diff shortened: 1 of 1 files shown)


Priyashetty17 (approved):
lgtm

extensions/vai/config.go
```diff
@@ -0,0 +1,9 @@
+package vai
+
+const (
+	ConfigurationFileKey = "vai"
```
anupama2501, line 4: is this constant going to be used pre dominantly in tests? if so, can  we move this to helper? If its going to b used outside of vai tests we can move this to actions..

brudnak:
Was thinking it could be useful for some tests to set in the config.

PR #225, opened 2024-06-26
update createCorral to run separately from wait & create to enable parallel corral jobs

slickwarren:
airgap cluster creates weren't working properly when specifying multiple pools with separate / mixed roles. This PR fixes that behavior by setting parity with the other cluster create functions, specifically using existing logic for selecting the right poolRole and quantity.

### DIFF
--- a/clients/corral/corral.go
+++ b/clients/corral/corral.go
@@ -4,22 +4,24 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"regexp"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/shepherd/pkg/session"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
-	debugFlag               = "--trace"
-	skipCleanupFlag         = "--skip-cleanup"
-	corralPrivateSSHKey     = "corral_private_key"
-	corralPublicSSHKey      = "corral_public_key"
-	corralRegistryIP        = "registry_ip"
-	corralRegistryPrivateIP = "registry_private_ip"
+	debugFlag           = "--trace"
+	skipCleanupFlag     = "--skip-cleanup"
+	corralPrivateSSHKey = "corral_private_key"
+	corralPublicSSHKey  = "corral_public_key"
 )
 
 // GetCorralEnvVar gets corral environment variables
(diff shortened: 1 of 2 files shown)


markusewalker (approved):

extensions/provisioning/creates.go
```diff
@@ -536,23 +536,29 @@ func CreateProvisioningRKE1CustomCluster(client *rancher.Client, externalNodePro
 // CreateProvisioningAirgapCustomCluster provisions a non-rke1 cluster using corral to gather its nodes, then runs verify checks
 func CreateProvisioningAirgapCustomCluster(client *rancher.Client, clustersConfig *clusters.ClusterConfig, corralPackages *corral.Packages) (*v1.SteveAPIObject, error) {
 	setLogrusFormatter()
-	rolesPerNode := map[int32]string{}
```
igomez06, line 539: Question why change this from a map? Isn't it much more efficient doing that rather than keeping track of two data structures?

slickwarren: this is matching what we have for all the other createCustomCluster functions. I just changed it to what I knew was working, since currently airgap wasn't working for multi-pool.

igomez06 (dismissed):
LGTM

igomez06:
Can we get back ports?

clients/corral/corral.go
```diff
@@ -68,9 +68,9 @@ func SetCustomRepo(repo string) error {
 	return nil
 }
 
-// CreateCorral creates a corral taking the corral name, the package path, and a debug set so if someone wants to view the
+// CreateCorralAndWait creates a corral taking the corral name, the package path, and a debug set so if someone wants to view the
 // corral create log
-func CreateCorral(ts *session.Session, corralName, packageName string, debug bool, cleanup bool) ([]byte, error) {
+func CreateCorralAndWait(ts *session.Session, corralName, packageName string, debug bool, cleanup bool) ([]byte, error) {
```
susesgartner, line 73: nit this can be shortened to`func CreateCorralAndWait(ts *session.Session, corralName, packageName string, debug, cleanup bool)`

clients/corral/corral.go
```diff
@@ -95,6 +94,42 @@ func CreateCorral(ts *session.Session, corralName, packageName string, debug boo
 	return msg, nil
 }
 
+// CreateCorral creates a corral taking the corral name, the package path, and a debug set so if someone wants to view the
+// corral create log
+func CreateCorral(ts *session.Session, corralName, packageName string, debug bool, cleanup bool) (*exec.Cmd, error) {
```
susesgartner, line 99: Same comment as above

clients/rancher/client.go
```diff
@@ -113,6 +115,14 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
...
+	splitBearerKey := strings.Split(bearerToken, ":")
+	token, err := c.Management.Token.ByID(splitBearerKey[0])
+	if err != nil {
+		return nil, err
+	}
+
+	c.UserID = token.UserID
+
 	return c, nil
 }
```
susesgartner, line 128: Unless i'm greatly mistaken these changes were from this [PR](https://github.com/rancher/shepherd/pull/250) and have already been merged.

susesgartner (commented):
Looks pretty good, just a few nits

clients/corral/corral.go
```diff
@@ -68,9 +68,9 @@ func SetCustomRepo(repo string) error {
 	return nil
 }
 
-// CreateCorral creates a corral taking the corral name, the package path, and a debug set so if someone wants to view the
+// CreateCorralAndWait creates a corral taking the corral name, the package path, and a debug set so if someone wants to view the
 // corral create log
-func CreateCorral(ts *session.Session, corralName, packageName string, debug bool, cleanup bool) ([]byte, error) {
+func CreateCorralAndWait(ts *session.Session, corralName, packageName string, debug, cleanup bool) ([]byte, error) {
```
igomez06, line 73: What's the point of this func? It's doing the exact same thing as the other CreateCorral.

slickwarren: createCorralAndWait is the old function, createCorral (if you look at shepherd / main)

CreateCorral does not use command().CombinedOutput (which waits on the job to finish, then returns) 
not using combinedOutput allows us to run corral in parallel, as now we don't have to wait on each job to finish in order to start the next one. 

see use of this here: https://github.com/rancher/shepherd/blob/782ae7375e7c01b7b9cc16d47e81b9cec54b6c84/extensions/provisioning/creates.go#L650-L655

igomez06: ah I see so you have removed CombinedOutput? 
Can you make this a separate PR? The provisioning has been moved anyways. Or you can wait until my work is done shouldn't be much longer.

slickwarren: yes, CombinedOutput is removed from the CreateCorral call, it is now in the CreateCorralAndWait call

slickwarren:
waiting on 2.9 actions v. extensions Rancher PR to go through (after code freeze) to fix the conflicts here.

slickwarren:
note; most of the changes for this original PR are now an action. The remainder is only for the corral change

bmdepesa:
A couple of questions:

- Is there a plan to add calls to `WaitOnCorralWithCombinedOutput` where `CreateCorral` is already being used in existing tests? 
- What happens if those calls aren't added, or someone forgets to call `WaitOnCorralWithCombinedOutput`? 
- Can the multithreaded/non-blocking creation and waiting be performed at a higher level so we don't need to specifically keep track of running commands?
- Have we looked into using waitgroups or similar functionality so we don't have to keep track of individual commands?

clients/corral/corral.go
```diff
@@ -95,6 +94,42 @@ func CreateCorral(ts *session.Session, corralName, packageName string, debug boo
...
+	}
+	args = append(args, corralName, packageName)
+	logrus.Infof("Creating corral with the following parameters: %v", args)
+
+	corralCommand := exec.Command("corral", args...)
+
+	var b bytes.Buffer
+	corralCommand.Stdout = &b
+	corralCommand.Stderr = &b
+	b.Bytes()
```
bmdepesa, line 119: This is unused. Can we remove it?

slickwarren: removed

bmdepesa:
One of the problems I keep coming back to is that we're breaking previous usage of `CreateCorral`. 

I'd suggest the following approach for discussion: 
- original CreateCorral -> remains unchanged
- new CreateCorral -> CreateCorralAsync
- WaitOnCorralWithCombinedOutput -> WaitForCommand (but moved to an extension as nothing about this is Corral-specific sans the error message)
    - errors.Wrap(err, "Unable to create corral: "+string(msg)) -> "Command '%s' had errors: %s", command, errors

Re: point 2 - I was more concerned about what happens at the goroutine level when we `Start()` a command but don't explicitly `Wait()` (not the behavioral aspect)

I'd advise looking into waitgroups for a more streamlined and less manual concurrency approach, but I think the suggested changes address most of my concern.

igomez06:
> One of the problems I keep coming back to is that we're breaking previous usage of `CreateCorral`.
> 
> I'd suggest the following approach for discussion:
> 
> * original CreateCorral -> remains unchanged
> * new CreateCorral -> CreateCorralAsync
> * WaitOnCorralWithCombinedOutput -> WaitForCommand (but moved to an extension as nothing about this is Corral-specific sans the error message)
>   
>   * errors.Wrap(err, "Unable to create corral: "+string(msg)) -> "Command '%s' had errors: %s", command, errors
> 
> Re: point 2 - I was more concerned about what happens at the goroutine level when we `Start()` a command but don't explicitly `Wait()` (not the behavioral aspect)
> 
> I'd advise looking into waitgroups for a more streamlined and less manual concurrency approach, but I think the suggested changes address most of my concern.

I 100% agree with this ideally we should probably use go routines and wait groups. But at least the separate async function addresses the concern of breaking existing functionality

slickwarren:
moving back to draft state until ^ above requests are addressed

slickwarren:
I'll get some jenkins jobs up for review on Monday, but I've tested this locally and it is working.

clients/corral/corral.go
```diff
@@ -70,29 +71,91 @@ func SetCustomRepo(repo string) error {
...
+func CreateCorral(ts *session.Session, corralName, packageName string, debug, cleanup bool) ([]byte, error) {
+	command, err := startCorral(ts, corralName, packageName, debug, false)
+	if err != nil {
+		return nil, err
+	}
+
+	return runAndWaitOnCommand(command)
+}
+
+func runAndWaitOnCommand(command *exec.Cmd) ([]byte, error) {
```
igomez06, line 83: With. doing it with the multiple wait groups is it necessary to abstract the run and wait command?

clients/corral/corral.go
```diff
@@ -70,29 +71,91 @@ func SetCustomRepo(repo string) error {
...
+		if err != nil {
+			errs = append(errs, err.Error())
+		}
+
+		waitGroup.Add(1)
+
+		go func() {
+			defer waitGroup.Done()
+
+			msg, err := runAndWaitOnCommand(cmdToRun)
```
igomez06, line 142: Why not just keep the CreateCorral the same, and have that command run in the go routine? There wouldn't be a need to just abstract the wait portion of it. This isn't exactly what we had in mind. It would be of

go func() {

   CreateCorral()
}


waitGroupWait()

clients/corral/corral.go
```diff
@@ -70,29 +74,151 @@ func SetCustomRepo(repo string) error {
...
+	err = waitForCorralConfig(corralName, packageName)
 	if err != nil {
-		return nil, errors.Wrap(err, "Unable to create corral: "+string(msg))
+		return nil, err
+	}
+
+	return cmdToRun, err
+}
+
+func waitForCorralConfig(corralName, packageName string) error {
```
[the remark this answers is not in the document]
slickwarren: the comment may be in the wrong place, but this is why 
```
	// this ensures corral is completely initiated. Otherwise, race conditions occur.
```

on line 130

basically yes, we're waiting for corral to put something in .corral/corrals/<package_name>, because that is its local cache of corral variables. If we don't wait for this to happen, then it opens up a race condition where .corral/config.yaml is updated by the next corral being started before the previous package's vars are cached in .corral/corrals/<package_name>

clients/corral/corral.go
```diff
@@ -70,29 +74,151 @@ func SetCustomRepo(repo string) error {
...
+
+	}
+
+	waitGroup.Wait()
+
+	var formattedError error
+	var longString string
+	if len(errStrings) > 0 {
+		for _, err := range errStrings {
+			logrus.Info(err)
```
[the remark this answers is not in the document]
slickwarren: I can remove this, it was for debugging purposes. The full errors should be included in the formattedError / longString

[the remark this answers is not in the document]
igomez06: yes please remove it

clients/corral/corral.go
caliskanugur: This still seems like the case, I can't find where is this `b` used, should it log it out info level?

clients/corral/corral.go
```diff
@@ -70,29 +74,150 @@ func SetCustomRepo(repo string) error {
...
+
+func waitForCorralConfig(corralName, packageName string) error {
+	backoff := wait.Backoff{
+		Duration: 1 * time.Second,
+		Factor:   1.1,
+		Jitter:   0.1,
+		Steps:    10,
+	}
+
+	corralOSPath := strings.Split(strings.ToLower(packageName), "/github")[0] + "/.corral/corrals/" + corralName + "/corral.yaml"
```
caliskanugur, line 147: Would this enforce cloning corral packages under github dir? If this is the case, would it make sense to use home path?

clients/corral/corral.go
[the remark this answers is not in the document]
caliskanugur: I've got a question here; I feel like this is partly happening because `exec.Start` and `exec.Wait` are separated. Did you have a chance to try the first iteration (unchanged version) of the `Create` corral within this `go func`? If that works, this may resolve what you are looking for without changing the existing `Create` and adding `startCorral`, `runAndWaitOnCommand` and `waitForCorralConfig`.

caliskanugur (commented):
Looks good, I've got some questions

clients/corral/corral.go
slickwarren: the path usually looks like `/root/go/src/github.com/rancherlabs/corral-packages/` in jenkins, but sometimes the string between `/root/` and `/github.com/` is different, so I found it best to lookup based on 'github'. Are you saying there's a home var I could use instead?

clients/corral/corral.go
igomez06: not this part though Caleb `b.Bytes()` this isn't doing anything. It's returning a bytes array you and your not setting it to any new variable https://pkg.go.dev/bytes#Buffer.Bytes

slickwarren: oh I see. Okay I can remove this line sorry I thought you meant this whole block

igomez06 (dismissed):
LGTM

clients/corral/corral.go
slickwarren: updated this to be based on the user's system

markusewalker (dismissed):

caliskanugur:
LGTM, do we need backports for this?

caliskanugur:
Also for this one's history, it looks like the rebase included the previous commits.

bmdepesa (approved):

igomez06 (approved):
LGTM

PR #227, opened 2024-07-03
Implemented time out report to log cluster status and condition infor…

lscalabrini01:
Implemented time out report to log cluster status and condition information to the provisioning tests

- Implemented TimeoutRKEReport function
- Implemented TimeoutClusterReport function
- Refactor watch function to use TimeoutError and WatchConnectionError variables
- Change VerifyRKE1Cluster to use TimeoutRKEReport function
- Change VerifyCluster to use TimeoutClusterReport function
- Change VerifyHostedCluster to use TimeoutRKEReport function

### DIFF
--- a/extensions/reports/timeout_report.go
+++ b/extensions/reports/timeout_report.go
@@ -1,0 +1,22 @@
+package reports
+
+import (
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
+	"github.com/rancher/shepherd/pkg/wait"
+	"github.com/sirupsen/logrus"
+)
+
+// TimeoutRKEReport is a function that print the State and Conditions of a RKE cluster if the error is a timeout error
+func TimeoutRKEReport(cluster *management.Cluster, err error) {
+	if err != nil && cluster != nil && err.Error() == wait.TimeoutError {
+		logrus.Errorf("Timeout report State: %s, Conditions %v", cluster.State, cluster.Conditions)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


extensions/reports/timeout_report.go
```diff
@@ -0,0 +1,13 @@
+package reports
+
+import (
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	"github.com/rancher/shepherd/pkg/wait"
+	"github.com/sirupsen/logrus"
+)
+
+func TimeoutReport(cluster *management.Cluster, err error) {
```
igomez06, line 9: Can you add go doc comment here?

igomez06 (commented):
looks great! Just one comment. And we can we get backports to release/v2.8 and release/v2.7?

extensions/reports/timeout_report.go
caliskanugur: Agreed,  also I have a question: will this work for v1 clusters?

caliskanugur (commented):
Looks awesome! just one question

extensions/reports/timeout_report.go
igomez06: Good point Ugur. You might need to have two functions. One for rke1 and one rke2/k3s

lscalabrini01: Yes, I can

lscalabrini01: I'll crate these new functions

floatingman (dismissed):
Looks good to me

floatingman (approved):
LGTM

PR #228, opened 2024-07-03
Fix controller generation to add test session to generated controllers

floatingman:
[2.8 Backport](https://github.com/rancher/shepherd/pull/237)

Refactor test session related functions in codegen

The code has been refactored to generalize the function of adding test sessions to different controllers. The functions for importing test session in each controller were renamed and refactored into a unified function, which is now called inside each controller-specific function. This approach aims to reduce code repetition and complexity.

Add session to struct and function in generated files

Expanded the appendSessionToStructInFile function to appendNewlineToBlockInFile, which generalizes line additions to the end of any struct or function block. This update improves code reusability. Utilized this function to add a session to multiple generated interface files.

Add session to struct in management controllers

A new function has been implemented to append a session to a struct in the management controllers. The function reads a file, locates a given struct, and then adds a new session prior to the closing brace of the struct. This change will be beneficial for the management of sessions within the struct.

Refactor import addition function for error handling

Updated the addImport function in codegen/main.go to handle and return errors instead of triggering panic. This change also includes verification of the import path and logic adjustments for adding a new import to the parsed file. These updates ensure more robust and error-resistant code execution.

Add 'session' import to controller files

This commit introduces additional functions to automatically add 'session' imports into controller files. Functions have been added to process and add this import into management, apps, and core controller files specifically. The new functions will walk the respective directory's file-paths and add the specified import to all 'interface.go' files.

Refactor factory method parameter replacement in codegen

The factory methods in the code generator have been refactored. The parameters passed into these methods have been modified to include the timestamp (TS) option. Also, excess white space in the code has been removed to adhere to clean coding principles.

### DIFF
--- a/pkg/codegen/main.go
+++ b/pkg/codegen/main.go
@@ -2,9 +2,17 @@ package main
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
+
+	"go/ast"
+	"go/parser"
+	"go/printer"
+	"go/token"
 
 	fleet "github.com/rancher/fleet/pkg/apis/fleet.cattle.io/v1alpha1"
 	"github.com/rancher/norman/types"
@@ -24,10 +32,21 @@ import (
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+// main initializes the code generation for controllers and clients.
+// It generates clients for various API groups using controllergen.Run.
+// It also generates clients for specific schemas using generator.GenerateClient.
+// Then, it calls replaceClientBasePackages to replace imports in the generated clients.
+// Finally, it replaces imports and adds controller test session for generated
 func main() {
 	err := os.Unsetenv("GOPATH")
 	if err != nil {
 		return
+	}
+
+	generatedControllerPaths := map[string]string{
+		"AppsControllerPath":       "./pkg/generated/controllers/apps",
+		"CoreControllerPath":       "./pkg/generated/controllers/core",
+		"ManagementControllerPath": "./pkg/generated/controllers/management.cattle.io",
 	}
 
 	controllergen.Run(args.Options{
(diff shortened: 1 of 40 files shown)


igomez06 (approved):
LGTM

pkg/codegen/main.go
```diff
@@ -250,3 +323,166 @@ func replaceImports(dir string) error {
...
+
+		// Write the modified content back to the file
+		err = os.WriteFile(filePath, []byte(modifiedContent), 0644)
+		if err != nil {
+			return fmt.Errorf("error writing to file: %v", err)
+		}
+		return nil
+	}
+
+	return nil
```
caliskanugur, line 487: Maybe I am reading this part wrong, but return nil will be the upper one always right? 
```suggestion
	}

	return nil
```

Should it rely on the return nil at the bottom? Would this cause a scope issue?

pkg/codegen/main.go
```diff
@@ -147,6 +155,19 @@ func main() {
 	if err := replaceCoreControllerImports(); err != nil {
 		panic(err)
 	}
+	// Comment out this function to avoid adding the test session import in the management controllers
```
caliskanugur, line 158: Definitely a nit question; maybe I don't get this bit, do we need to comment out the lines? Files are going to be deleted and this will need to happen on the newly generated files right?

caliskanugur (commented):
Looks great! Just have a couple of questions

pkg/codegen/main.go
```diff
@@ -187,7 +208,50 @@ func replaceCoreControllerImports() error {
...
+// Walk through the generated core controllers and add test session
+// to necessary functions and structs
+func addCoreControllerTestSession() error {
+	root := "./pkg/generated/controllers/core"
+	err := filepath.Walk(root, processInterfaceFile)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
igomez06: That's a good point. Would it make sense to have a generic function, and keep track of the paths as const strings?

pkg/codegen/main.go
floatingman: The other team started this comment pattern when they introduced the replaceImport function. It can be taken out. It doesn't matter much to me.

pkg/codegen/main.go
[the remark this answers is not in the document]
floatingman: done

pkg/codegen/main.go
floatingman: Fixed

floatingman:
Made requested changes and added go fmt to generate cmd.
The majority of the file changes just deal with correcting the formatting. Only main.go in the codegen package contains the logic and function updates.

igomez06 (approved):
LGTM

caliskanugur (approved):
Looks good to me

PR #231, opened 2024-07-11
[v2.10] Update LoadUpgradeKubernetesConfig for upgrade refactor

markusewalker:
### PR Description
As part of the QA task, we are refactoring the upgrading package to follow the best practices that packages such as `provisioning` and `snapshot` follow. When refactoring, it was agreed upon to have `kubernetes_test.go` perform the pre and post upgrade checks instead of `workload_test.go`. That way, one test runs all of the operations.

How it currently is, you have to run 3 individual tests that all pertain to one Kubernetes upgrade. This is being addressed in the r/r PR. For shepherd, we need to update the `LoadUpgradeKubernetesConfig` to ensure that the parameters found in `LoadUpgradeWorkloadsConfig` are being referenced.

This way, the `kubernetes_test.go` does not need to have to call two separate configs, which is making the refactor overly complicated.

### DIFF
--- a/extensions/clusters/clusters.go
+++ b/extensions/clusters/clusters.go
@@ -10,9 +10,11 @@ import (
 	"github.com/rancher/norman/types"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/shepherd/clients/rancher"
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/defaults"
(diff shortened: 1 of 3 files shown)


Josh-Diamond (approved):

Josh-Diamond (approved):

extensions/clusters/clusters.go
```diff
@@ -1060,6 +1228,38 @@ func DeleteK3SRKE2Cluster(client *rancher.Client, clusterID string) error {
 	return nil
 }
 
+// UpdateRKE1Cluster is a "helper" functions that takes a rancher client, old rke1 cluster config, and the new rke1 cluster config as parameters.
+func UpdateRKE1Cluster(client *rancher.Client, cluster, updatedCluster *management.Cluster) (*management.Cluster, error) {
```
slickwarren, line 1232: I see that this was here for rke2/k3s already but IMO these helpers shouldn't live in clusters.go. We were intentionally not putting functions that `wait` in clusters.go, as that is part of verify.

markusewalker: Heard. Personally, I think that's a task that's outside the scope of this. I would want to then move all of the functions that do that into verify, not just this new function.

slickwarren (approved):

extensions/clusters/clusters.go
slickwarren: would you mind making a task for this?

markusewalker: Yep, made over here: https://github.com/rancher/qa-tasks/issues/1431

PR #232, opened 2024-07-13
update Output() to return map[string]any instead of []map[string]any

git-ival:
Small update to the return value of the tfexec.Output() function


igomez06 (approved):
LGTM we need back ports correct?

PR #233, opened 2024-07-19
Credentials update

susesgartner:
Update cloud credentials functions for every provider.

### DIFF
--- a/extensions/cloudcredentials/aws/create.go
+++ b/extensions/cloudcredentials/aws/create.go
@@ -2,28 +2,43 @@ package aws
 
 import (
 	"github.com/rancher/shepherd/clients/rancher"
-	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/cloudcredentials"
-	"github.com/rancher/shepherd/pkg/config"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/defaults/namespaces"
+	"github.com/rancher/shepherd/extensions/defaults/providers"
+	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
+	"github.com/rancher/shepherd/extensions/steve"
+	"github.com/rancher/shepherd/pkg/namegenerator"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const awsCloudCredNameBase = "awsCloudCredential"
-
-// CreateAWSCloudCredentials is a helper function that takes the rancher Client as a parameter and creates
-// an AWS cloud credential, and returns the CloudCredential response
-func CreateAWSCloudCredentials(rancherClient *rancher.Client) (*cloudcredentials.CloudCredential, error) {
-	var amazonEC2CredentialConfig cloudcredentials.AmazonEC2CredentialConfig
-	config.LoadConfig(cloudcredentials.AmazonEC2CredentialConfigurationFileKey, &amazonEC2CredentialConfig)
-
-	cloudCredential := cloudcredentials.CloudCredential{
-		Name:                      awsCloudCredNameBase,
-		AmazonEC2CredentialConfig: &amazonEC2CredentialConfig,
+// CreateAWSCloudCredentials is a helper function that creates V1 cloud credentials and waits for them to become active.
+func CreateAWSCloudCredentials(client *rancher.Client, credentials cloudcredentials.CloudCredential) (*v1.SteveAPIObject, error) {
+	secretName := namegenerator.AppendRandomString(providers.AWS)
+	spec := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: cloudcredentials.GeneratedName,
+			Namespace:    namespaces.CattleData,
+			Annotations: map[string]string{
+				"provisioning.cattle.io/driver": providers.AWS,
+				"field.cattle.io/name":          secretName,
+				"field.cattle.io/creatorId":     client.UserID,
+			},
+		},
+		Data: map[string][]byte{
+			"amazonec2credentialConfig-accessKey":     []byte(credentials.AmazonEC2CredentialConfig.AccessKey),
+			"amazonec2credentialConfig-secretKey":     []byte(credentials.AmazonEC2CredentialConfig.SecretKey),
+			"amazonec2credentialConfig-defaultRegion": []byte(credentials.AmazonEC2CredentialConfig.DefaultRegion),
+		},
+		Type: corev1.SecretTypeOpaque,
 	}
 
-	resp := &cloudcredentials.CloudCredential{}
-	err := rancherClient.Management.APIBaseClient.Ops.DoCreate(management.CloudCredentialType, cloudCredential, resp)
+	awsCloudCredentials, err := steve.CreateAndWaitForResource(client, stevetypes.Secret, spec, true, defaults.FiveSecondTimeout, defaults.FiveMinuteTimeout)
 	if err != nil {
 		return nil, err
 	}
-	return resp, nil
+
+	return awsCloudCredentials, nil
 }
--- a/extensions/cloudcredentials/azure/create.go
+++ b/extensions/cloudcredentials/azure/create.go
@@ -2,28 +2,44 @@ package azure
 
 import (
 	"github.com/rancher/shepherd/clients/rancher"
-	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/cloudcredentials"
-	"github.com/rancher/shepherd/pkg/config"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/defaults/namespaces"
+	"github.com/rancher/shepherd/extensions/defaults/providers"
+	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
+	"github.com/rancher/shepherd/extensions/steve"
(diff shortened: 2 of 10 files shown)


extensions/machinepools/digitalocean_machine_config.go
```diff
@@ -48,7 +49,7 @@ func NewDigitalOceanMachineConfig(generatedPoolName, namespace string) []unstruc
 		machineConfig.SetGenerateName(generatedPoolName)
 		machineConfig.SetNamespace(namespace)
 
-		machineConfig.Object["accessToken"] = ""
+		machineConfig.Object["accessToken"] = doMachineConfig.AccessToken
```
slickwarren, line 52: can you double check this, I'm pretty sure it is supposed to be blank for the machine / node templates. Otherwise, you'd need to update the other files in this folder that set it to blank for say linode.

extensions/cloudcredentials/aws/create.go
```diff
@@ -1,29 +1,53 @@
...
+		return nil, err
+	}
 
-	cloudCredential := cloudcredentials.CloudCredential{
-		Name:                      awsCloudCredNameBase,
-		AmazonEC2CredentialConfig: &amazonEC2CredentialConfig,
+	secretName := namegenerator.AppendRandomString(awsProvider)
+	spec := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "cc-",
```
slickwarren, line 31: if there's a way to make this global for all the cloudcredentials that would be nice.

slickwarren (commented):
also ping'd offline to resolve something

extensions/cloudcredentials/aws/create.go
susesgartner: Updated!

extensions/cloudcredentials/aws/create.go
```diff
@@ -1,29 +1,53 @@
...
+			Namespace:    credentialNamespace,
+			Annotations: map[string]string{
+				"provisioning.cattle.io/driver": awsProvider,
+				"field.cattle.io/name":          secretName,
+				"field.cattle.io/creatorId":     users.Data[0].ID,
```
slickwarren, line 36: is the 0th user always the one you expect? In the past, we've run into trouble with how norman sorts things

extensions/cloudcredentials/digitalocean/create.go
```diff
@@ -1,29 +1,51 @@
...
+			GenerateName: cloudcredentials.GeneratedName,
+			Namespace:    credentialNamespace,
+			Annotations: map[string]string{
+				"field.cattle.io/name":          secretName,
+				"provisioning.cattle.io/driver": digitalOceanProvider,
+				"field.cattle.io/creatorId":     users.Data[0].ID,
+			},
+		},
+		Data: map[string][]byte{
+			"digitalOceancredentialConfig-authEncodedJson": []byte(credentials.DigitalOceanCredentialConfig.AccessToken),
```
slickwarren, line 40: confirming this name is correct; since DO uses a token, not json afaik

extensions/rke1/nodetemplates/aws/create.go
```diff
@@ -16,7 +17,9 @@ func CreateAWSNodeTemplate(rancherClient *rancher.Client) (*nodetemplates.NodeTe
 	var amazonEC2NodeTemplateConfig nodetemplates.AmazonEC2NodeTemplateConfig
 	config.LoadConfig(nodetemplates.AmazonEC2NodeTemplateConfigurationFileKey, &amazonEC2NodeTemplateConfig)
 
-	cloudCredential, err := aws.CreateAWSCloudCredentials(rancherClient)
+	var cloudCredentialConfig cloudcredentials.CloudCredential
+	config.LoadConfig(cloudcredentials.AmazonEC2CredentialConfigurationFileKey, &cloudCredentialConfig.AmazonEC2CredentialConfig)
```
slickwarren, line 21: why is the fileKey loaded 2x?

slickwarren (commented):
a few nits; waiting for jenkins runs for approval

extensions/cloudcredentials/azure/create.go
```diff
@@ -1,29 +1,54 @@
...
-		AzureCredentialConfig: &azureCredentialConfig,
+	secretName := namegenerator.AppendRandomString(azureProvider)
+	spec := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: cloudcredentials.GeneratedName,
+			Namespace:    credentialNamespace,
+			Annotations: map[string]string{
+				"provisioning.cattle.io/driver": azureProvider,
+				"field.cattle.io/name":          secretName,
+				"field.cattle.io/creatorId":     users.Data[0].ID,
```
markusewalker, line 36: Same comment as the AWS above.

extensions/cloudcredentials/cloudcredentials.go
```diff
@@ -23,3 +31,54 @@ type CloudCredential struct {
 	VmwareVsphereConfig          *VmwarevsphereCredentialConfig `json:"vmwarevspherecredentialConfig,omitempty"`
 	UUID                         string                         `json:"uuid,omitempty"`
 }
+
+func LoadCloudCredential(provider string) CloudCredential {
```
markusewalker, line 35: nit - can you add a godoc here?

extensions/cloudcredentials/cloudcredentials.go
```diff
@@ -23,3 +31,54 @@ type CloudCredential struct {
...
+	var cloudCredential CloudCredential
+	switch {
+
+	case provider == provisioninginput.AWSProviderName.String():
+		var awsCredentialConfig AmazonEC2CredentialConfig
+		config.LoadConfig(AmazonEC2CredentialConfigurationFileKey, &awsCredentialConfig)
+		cloudCredential.AmazonEC2CredentialConfig = &awsCredentialConfig
+		return cloudCredential
```
markusewalker, line 43: nit - can you add a new line  after 42? So that the return is not so consolidated? Same for the rest in this switch statement.

extensions/cloudcredentials/cloudcredentials.go
```diff
@@ -23,3 +31,54 @@ type CloudCredential struct {
...
+
+	case provider == provisioninginput.GoogleProviderName.String():
+		var googleCredentialConfig GoogleCredentialConfig
+		config.LoadConfig(GoogleCredentialConfigurationFileKey, &googleCredentialConfig)
+		cloudCredential.GoogleCredentialConfig = &googleCredentialConfig
+		return cloudCredential
+
+	default:
+		panic(fmt.Sprintf("Provider:%v not found", provider))
```
markusewalker, line 82: Is there a specific reason we are `panic`? Couldn't we just do a `logrus.Errorf` or something and not panic?

markusewalker (commented):
Left some comments, will wait for new jobs to kick off.

extensions/cloudcredentials/aws/create.go
```diff
@@ -1,29 +1,53 @@
...
+)
 
-// CreateAWSCloudCredentials is a helper function that takes the rancher Client as a parameter and creates
-// an AWS cloud credential, and returns the CloudCredential response
-func CreateAWSCloudCredentials(rancherClient *rancher.Client) (*cloudcredentials.CloudCredential, error) {
-	var amazonEC2CredentialConfig cloudcredentials.AmazonEC2CredentialConfig
-	config.LoadConfig(cloudcredentials.AmazonEC2CredentialConfigurationFileKey, &amazonEC2CredentialConfig)
+// CreateAWSCloudCredentials is a helper function that creates V1 cloud credentials and waits for them to become active.
+func CreateAWSCloudCredentials(client *rancher.Client, credentials cloudcredentials.CloudCredential) (*v1.SteveAPIObject, error) {
+	users, err := client.Management.User.ListAll(&types.ListOpts{})
```
igomez06, line 23: I agree with Markus and Caleb, getting the full list of users and then assuming it's the first entry sounds like would break when multiple users have been created. Perhaps we pass in the user data as well?

extensions/cloudcredentials/cloudcredentials.go
igomez06: I think Sam is good here, if the user provides a non-supported value we should panic.

igomez06 (commented):
overall looks good, I have the same concerns as Markus and Caleb. Would definitely need see a run with all providers to provide a sign off.

extensions/cloudcredentials/aws/create.go
susesgartner: It was in the case of admin/standard clients. That being said I do wonder if there is a better way to pull this ID without

extensions/rke1/nodetemplates/aws/create.go
susesgartner: The first load is for the nodeTemplateConfig, the second is for the Credentials.

extensions/cloudcredentials/aws/create.go
igomez06: I think we should pass the user ID instead. How do we guarantee that the first element of the user's list is the user that created the credential?

igomez06 (changes requested):
Better way of getting the Creator ID

extensions/cloudcredentials/aws/create.go
susesgartner: This has been fixed!

markusewalker (dismissed):

igomez06 (approved):
Looks great.

igomez06:
can we get back ports too?

susesgartner:
I'll add the backports once code freeze is lifted.

caliskanugur (approved):
Looks good to me

markusewalker:
Thanks @susesgartner ! If possible, please hold off on merging until the freeze is lifted and the backports can be merged alongside this.

slickwarren (approved):

caliskanugur:
Backports:

- https://github.com/rancher/shepherd/pull/280
- https://github.com/rancher/shepherd/pull/281
- https://github.com/rancher/shepherd/pull/282

markusewalker (approved):

PR #243, opened 2024-07-30
Refactor context to store session for downstream context creation

Priyashetty17:
https://github.com/rancher/qa-tasks/issues/1443

**Issue**:
The DownStreamClusterWranglerContext function is currently not handling sessions correctly. When this function is called, it passes nil for the session parameter, resulting in a panic when objects are created on the downstream cluster using wrangler client.

**Solution**:

- Add a session attribute to the Context struct
- Update the NewContext function to initialize the session attribute
- Modify DownStreamClusterWranglerContext to use the stored session

**2.8 Backport PR**: https://github.com/rancher/shepherd/pull/244

### DIFF
--- a/pkg/wrangler/context.go
+++ b/pkg/wrangler/context.go
@@ -81,6 +81,7 @@ type Context struct {
 	apps *apps.Factory
 	core *core.Factory
 
+	session *session.Session
 	started bool
 }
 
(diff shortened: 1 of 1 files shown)


caliskanugur:
It looks good to me, @Priyashetty17 could you confirm this is fixing the existing problem and session with downstream is working as intended and queueing the created downstream resources via the wrangler context?

Priyashetty17:
@caliskanugur  Yes, the tests are passing now. I have shared the freeform job with you on slack.

caliskanugur (approved):
LGTM

PR #249, opened 2024-08-05
[Main] Add Windows support for scaling nodepools

markusewalker:
### PR Description
Small PR that adds a Windows conditional check when scaling RKE2 machinepools up and down. This is applicable only for vSphere node driver clusters that have a Windows worker(s).

### DIFF
--- a/extensions/workloads/template.go
+++ b/extensions/workloads/template.go
@@ -24,9 +24,13 @@ func NewContainer(containerName, image string, imagePullPolicy corev1.PullPolicy
 }
 
 // NewPodTemplate is a constructor that creates the pod template for all types of workloads e.g. cronjobs, daemonsets, deployments, and batch jobs
-func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imagePullSecrets []corev1.LocalObjectReference, labels map[string]string) corev1.PodTemplateSpec {
+func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imagePullSecrets []corev1.LocalObjectReference, labels map[string]string, nodeSelector map[string]string) corev1.PodTemplateSpec {
 	if labels == nil {
 		labels = make(map[string]string)
+	}
+
+	if nodeSelector == nil {
+		nodeSelector = make(map[string]string)
 	}
 
 	return corev1.PodTemplateSpec{
@@ -37,6 +41,7 @@ func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imag
 			Containers:       containers,
 			Volumes:          volumes,
 			ImagePullSecrets: imagePullSecrets,
+			NodeSelector:     nodeSelector,
 		},
 	}
 }
@@ -67,7 +72,7 @@ func NewDeploymentTemplate(deploymentName string, namespace string, template cor
 
 }
 
-// NewDeploymentTemplate is a constructor that creates a daemonset template. If the isCattleLabeled true, workloadselector labels are assigned to the daemonset and the pod template.
+// NewDaemonSetTemplate is a constructor that creates a daemonset template. If the isCattleLabeled true, workloadselector labels are assigned to the daemonset and the pod template.
 func NewDaemonSetTemplate(daemonsetName string, namespace string, template corev1.PodTemplateSpec, isCattleLabeled bool, matchLabels map[string]string) *appv1.DaemonSet {
 	if matchLabels == nil {
 		matchLabels = map[string]string{}


susesgartner (approved):
LGTM

extensions/machinepools/machinepools.go
```diff
@@ -54,6 +56,10 @@ func MatchNodeRolesToMachinePool(nodeRoles NodeRoles, machinePools []apisV1.RKEM
 		return index, count
 	}
 
+	if nodeRoles.Windows {
+		return len(machinePools) - 1, count
```
slickwarren, line 60: what if I didn't put windows as the last machinePool though?

markusewalker: Is there a reason why you wouldn't, though? I know vSphere allows Windows with Linux out the gate, but with typical K8s clusters, Windows has to come after a cluster is already active with Linux nodes. I would expect we follow the same standard where Windows comes last.

slickwarren: rancher / the framework handles adding that last for the user iirc. Generally its at the end of the list, but there's also the case if I have more than 1 machinePool with windows nodes with custom quantities.

Is there a way to look it up instead?

markusewalker: Addressed. There is an enhancement that I will make in a separate PR to account for multiple machine pools with the same role. That's a bit outside the scope of this PR.

extensions/machinepools/machinepools.go
```diff
@@ -54,6 +58,14 @@ func MatchNodeRolesToMachinePool(nodeRoles NodeRoles, machinePools []apisV1.RKEM
 		return index, count
 	}
 
+	if nodeRoles.Windows {
+		for index, machinePoolConfig := range machinePools {
```
slickwarren, line 62: can you just add a comment as to why there's a for loop so we know why there's 2 in the function, otherwise lgtm

markusewalker: Added a comment here 👍🏿

markusewalker:
With Israel's upcoming actions PR, this file is actually being moved back to `rancher/rancher`. When that PR gets merged, I will close this PR as it is no longer relevant.

markusewalker:
With Israel's upcoming actions PR, this file is actually being moved back to rancher/rancher. When that PR gets merged, I will close this PR as it is no longer relevant.

Josh-Diamond (dismissed):

extensions/workloads/template.go
```diff
@@ -42,7 +42,7 @@ func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imag
 }
 
 // NewDeploymentTemplate is a constructor that creates a deployment template. If the isCattleLabeled true, workloadselector labels are assigned to the deployment and the pod template.
-func NewDeploymentTemplate(deploymentName string, namespace string, template corev1.PodTemplateSpec, isCattleLabeled bool, matchLabels map[string]string) *appv1.Deployment {
+func NewDeploymentTemplate(deploymentName string, namespace string, template corev1.PodTemplateSpec, isCattleLabeled bool, matchLabels map[string]string, isWindows bool) *appv1.Deployment {
```
slickwarren, line 45: since `template` is an input variable, wouldn't it make sense to add the Spec.NodeSelector before calling this NewDeploymentTemplate function? i.e. do the if isWindows statement in the NewPodTemplate function instead of here.

markusewalker: I personally think it makes more sense to leave it as it is. When I did my comprehensive testing on my final commit and during release testing, multiple permutations for clusters had to be done. And having it at this level caused no issues in any of the cluster configurations.

igomez06: I 100% agree with Caleb, doesn't it make way more since to do this in NewPodTemplate instead? template is an input variable. Also make sure the constructor is appropriately updated in tests/actions that use this function

extensions/workloads/template.go
```diff
@@ -78,6 +86,14 @@ func NewDaemonSetTemplate(daemonsetName string, namespace string, template corev
 		template.ObjectMeta.Labels["workload.user.cattle.io/workloadselector"] = fmt.Sprintf("apps.daemonset-%v-%v", namespace, daemonsetName)
 	}
 
+	if isWindows {
```
igomez06, line 89: Yeah let's add this in NewPodTemplate that way it's a single if statement rather than affecting two functions for this.

extensions/workloads/template.go
markusewalker: Done.

extensions/workloads/template.go
markusewalker: Done.

markusewalker: I actually need this here. This specifically is causing upgrade tests to fail without it. It's because Linux and Windows worker nodes are being registered, so there needs to be the filter to properly have this.

slickwarren: Can you just move this if statement to the NewPodTemplate?

markusewalker: I can try tomorrow. I actually used a wrong label when I was doing local testing lol. Will follow up tomorrow.

markusewalker: That worked, I did a quick local test with snapshots / upgrades and validated all was good.

extensions/workloads/template.go
```diff
@@ -24,9 +24,18 @@ func NewContainer(containerName, image string, imagePullPolicy corev1.PullPolicy
 }
 
 // NewPodTemplate is a constructor that creates the pod template for all types of workloads e.g. cronjobs, daemonsets, deployments, and batch jobs
-func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imagePullSecrets []corev1.LocalObjectReference, labels map[string]string) corev1.PodTemplateSpec {
+func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imagePullSecrets []corev1.LocalObjectReference, labels map[string]string, isWindows bool) corev1.PodTemplateSpec {
+	var nodeSelector map[string]string
 	if labels == nil {
 		labels = make(map[string]string)
+
+		if isWindows {
```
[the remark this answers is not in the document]
markusewalker: Spoke offline, resolved.

extensions/workloads/template.go
```diff
@@ -37,6 +46,7 @@ func NewPodTemplate(containers []corev1.Container, volumes []corev1.Volume, imag
 			Containers:       containers,
 			Volumes:          volumes,
 			ImagePullSecrets: imagePullSecrets,
+			NodeSelector:     nodeSelector,
```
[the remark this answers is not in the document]
markusewalker: Spoke offline, resolved.

slickwarren (approved):
lgtm

Josh-Diamond (approved):

bmdepesa (changes requested):
Conditional logic in our frameworks/tests is not the best design... and I'm not actually sure why this method exists to begin with since it's just returning a PodTemplateSpec (but that should be a separate discussion).

A better approach would be to remove the `isWindows` param and replace with a nodeSelector parameter and have that handled as the labels currently are. 

The labels/nodeselectors should then be passed in from where they're being used in the (hopefully Windows-specific) tests.

Josh-Diamond (dismissed):

bmdepesa (approved):

Josh-Diamond (approved):

slickwarren (approved):

susesgartner (approved):

igomez06 (approved):
LGTM

PR #250, opened 2024-08-05
Add the user field to rancher clients.

susesgartner:
Purpose: To add the user to the client so functions have an easy way to access the user the client that created the client without having to pass that user from the test layer down to the function using it (which can be a headache)

### DIFF
--- a/clients/rancher/client.go
+++ b/clients/rancher/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/norman/httperror"
@@ -52,6 +53,7 @@ type Client struct {
 	// Flags is the environment flags used by the client to test selectively against a rancher instance.
 	Flags      *environmentflag.EnvironmentFlags
 	restConfig *rest.Config
+	UserID     string
 }
 
 // NewClient is the constructor to the initializing a rancher Client. It takes a bearer token and session.Session. If bearer token is not provided,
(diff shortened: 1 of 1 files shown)


clients/rancher/client.go
```diff
@@ -113,6 +115,17 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 
 	c.WranglerContext = wranglerContext
 
+	splitBearerKey := strings.Split(bearerToken, ":")
+	token, err := c.Management.Token.ByID(splitBearerKey[0])
+	if err != nil {
+		return nil, err
+	}
+
+	c.User, err = c.Management.User.ByID(token.UserID)
```
igomez06, line 124: I don't think we should get the user? I think we should just have the user ID as the client attribute and if any call needs anything extra it can get the user. That way we limit our instantiation to one GET call.

susesgartner: Ok, I'll make the change.

clients/rancher/client.go
```diff
@@ -52,6 +53,7 @@ type Client struct {
 	// Flags is the environment flags used by the client to test selectively against a rancher instance.
 	Flags      *environmentflag.EnvironmentFlags
 	restConfig *rest.Config
+	CreatorID  string
```
caliskanugur, line 56: Should we call this UserID? 

Especially considering it'd look like this: `client.UserID`

Don't have strong opinions on naming though

caliskanugur (commented):
Looks good, a net improvement for Shepherd

The only caveat here is adding another GET request in the client instantiation.

Could you import this to r/r main and observe r/r integration tests workflow multiple times by rerunning it?

clients/rancher/client.go
susesgartner: I have it as CreatorID simply because thats the naming convention that is used in rancher for a variety of objects. Figured i'd stick with the theme, if your ok with that.

igomez06:
We just need it to pass the GHA on rancher and few times, and we should be good to go.

susesgartner:
Sounds good I just made a rancher PR for it. I'll rerun it a few times after it passes.

clients/rancher/client.go
susesgartner: Updated it!

markusewalker (approved):

markusewalker:
Can we get backports for this @susesgartner?

caliskanugur (approved):
LGTM

PR #253, opened 2024-08-06
Updating the workflow to skip draft PRs

anupama2501:
https://github.com/rancher/qa-tasks/issues/1259

- Excludes draft PRs after obtaining reviews.
- Changed the sha of the actions/checkout and actions/github-scripts to the latest release as there were some warnings in the console logs while the action runs.
- Added bold formatting to the report statements.
 
Report will be sent on slack.

### DIFF
--- a/.github/workflows/stale-prs.yml
+++ b/.github/workflows/stale-prs.yml
@@ -12,11 +12,11 @@ jobs:
 
     steps:      
       - name: Checkout code
-        uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
         
       - name: Check PRs and compile report
         id: compile_report
-        uses: actions/github-script@211cb3fefb35a799baa5156f9321bb774fe56294
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
         with:
           script: |
             const fiveDaysAgo = new Date(new Date().setDate(new Date().getDate() - 5)); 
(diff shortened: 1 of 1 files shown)


rancher-max (dismissed):
Niiicee I like the formatting updates. Looks clean!

.github/workflows/stale-prs.yml
```diff
@@ -51,8 +51,13 @@ jobs:
                     per_page: 100,
                   });
                   
+                  if (pr.draft) {
+                    // debugging
+                    console.log(`Skipping draft PR: ${pr.title}`);
```
caliskanugur, line 56: This part is from dev stage

.github/workflows/stale-prs.yml
```diff
@@ -51,8 +51,13 @@ jobs:
                     per_page: 100,
                   });
                   
+                  if (pr.draft) {
```
caliskanugur, line 54: Should it cover undefined chance here or are we good to accept `pr` and `pr.draft` are not undefined at this point?

caliskanugur (commented):
Looks good, couple comments

.github/workflows/stale-prs.yml
slickwarren: agreed, log can be rm'd

slickwarren (commented):
overall lgtm, one small nit

.github/workflows/stale-prs.yml
anupama2501: github docs state `Draft pull requests are available in public repositories with GitHub Free and GitHub Free for organizations, GitHub Pro, and legacy per-repository billing` https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests

So pr.draft should be not undefined at this point. and for pr also at this point it should not be undefined because it would have failed at line 39 itself. we can add a check to verify pr is not undefined at an earlier point tho.

floatingman (approved):
LGTM

slickwarren (approved):

PR #256, opened 2024-08-12
updating sortReleases key=minor version instead of rancherMinorversion

slickwarren:


extensions/clusters/kubernetesversions/all.go
caliskanugur, line 1: Could you also update the sorting logic comment at the top of the function? I believe the second point isn't applicable anymore

extensions/clusters/kubernetesversions/all.go
```diff
@@ -215,23 +215,19 @@ func sortReleases(releases []interface{}) (allAvailableVersions []string) {
...
-			continue
-		}
-
 		releaseKubernetesVersion, err := semver.NewVersion(strings.TrimPrefix(kubernetesVersion, "v"))
 		if err != nil {
 			continue
 		}
 
-		value, ok := availableVersionsMap[minRancherVersion.String()]
+		minor := strings.Split(releaseKubernetesVersion.Original(), ".")[1]
```
anupama2501, line 225: qq, what does releaseKubernetesVersion.Original() return here? can we not just use releaseKubernetesVersion?

slickwarren: .Original returns the string of releaseKuberenetesVersion, which is of type semver.Version. So we need the conversion to string here

extensions/clusters/kubernetesversions/all.go
slickwarren: updated the comment to be more accurate

caliskanugur (approved):
Looks good to me

floatingman (approved):
LGTM

anupama2501 (approved):

markusewalker (approved):

PR #257, opened 2024-08-13
[Main] Moved Actions.

igomez06:
Moved the identified extensions that should be actions out, and into rancher/rancher.


markusewalker (approved):
LGTM. Any possible regressions can be fixed as they're found in testing specific packages.

slickwarren:
Met with Israel and here are the rules to further help distinguish between actions v. Extensions. Please use this for reference when reviewing:


The following qualify a function to be an extension:
* Must be either an api call not natively captured by the client OR specific CRUD / wait on resource
  * i.e. download ssh keys
  * tokens package
  * should not only directly convert -> this would make it an action

Disqualifiers for extensions (If any one or more are true, function is an action):
* if there's a custom config needed, in any part
* if it is a validation of any sort (Waits are okay)
* if function is a direct conversion of a resource
  * more on this, if the function is purely just calling a native call + catching the error, this should not be an extension. A conversion would be more complex; see an example like token package
* how re-usable is the code? Shepherd code should be highly re-usable

slickwarren (commented):
most comments are in the rancher/rancher PR since they are complementary of eachother. But I have some questions:
* I think harden would be an action. It is dependent on the custom config for NodeRoles and ties very closely with that package. wdyt?
* I think the kubeconfig package is too specific and don't see it being used much outside of very specific rancher areas; this could be an action, I think.

igomez06:
> most comments are in the rancher/rancher PR since they are complementary of eachother. But I have some questions:
> 
> * I think harden would be an action. It is dependent on the custom config for NodeRoles and ties very closely with that package. wdyt?
> * I think the kubeconfig package is too specific and don't see it being used much outside of very specific rancher areas; this could be an action, I think.

The harden functions that were clusters.go did get move. The harden package is a lot more generic and I think can stay. Is that what you meant?

I disagree about the kubeconfig package. Look at it closely these are kubernetes specific stuff. kubectl, I can see your point, I think this package should be refactored to use the command line kubectl possibly making it a client instead.

caliskanugur:
> > most comments are in the rancher/rancher PR since they are complementary of eachother. But I have some questions:
> > 
> > * I think harden would be an action. It is dependent on the custom config for NodeRoles and ties very closely with that package. wdyt?
> > * I think the kubeconfig package is too specific and don't see it being used much outside of very specific rancher areas; this could be an action, I think.
> 
> The harden functions that were clusters.go did get moved. The harden package is a lot more generic and I think can stay. Is that what you meant?
> 
> I disagree about the kubeconfig package. Look at it closely these are kubernetes specific stuff. kubectl, I can see your point, I think this package should be refactored to use the command line kubectl possibly making it a client instead. But in these are generic funcs to run kubectl on rancher downstream clusters.

I think the hardening functions are good candidates to be an extension but not in their current state. The parameters are customized to provisioning create functions and are strongly tied. Though what it does in the function is more like an extension, in my opinion, they're still actions

For kubeconfig, I agree they are extensions, they are generic Kubernetes utilities that could be used across repos.

igomez06:
> > > most comments are in the rancher/rancher PR since they are complementary of eachother. But I have some questions:
> > > 
> > > * I think harden would be an action. It is dependent on the custom config for NodeRoles and ties very closely with that package. wdyt?
> > > * I think the kubeconfig package is too specific and don't see it being used much outside of very specific rancher areas; this could be an action, I think.
> > 
> > 
> > The harden functions that were clusters.go did get moved. The harden package is a lot more generic and I think can stay. Is that what you meant?
> > I disagree about the kubeconfig package. Look at it closely these are kubernetes specific stuff. kubectl, I can see your point, I think this package should be refactored to use the command line kubectl possibly making it a client instead. But in these are generic funcs to run kubectl on rancher downstream clusters.
> 
> I think the hardening functions are good candidates to be extensions but not in their current state. The parameters are customized to provisioning create functions and are strongly tied. Though what it does in the function is more like an extension, in my opinion, they're still actions
> 
> For kubeconfig, I agree they are extensions, they are generic Kubernetes utilities that could be used across repos.

Ah i see both of your points. I'll move it. Thanks!

caliskanugur (approved):
Looks good to me

igomez06:
backports: 

- https://github.com/rancher/shepherd/pull/262
- https://github.com/rancher/shepherd/pull/263
- https://github.com/rancher/shepherd/pull/264

PR #263, opened 2024-08-19
[v2.8] Moved actions

igomez06:
https://github.com/rancher/qa-tasks/issues/1300
Moved the identified extensions that should be actions out, and into rancher/rancher.


caliskanugur (approved):
Looks good to me

markusewalker (approved):

PR #267, opened 2024-08-22
<2.7> Fixed pod logs streaming error

lscalabrini01:
Changed kubectl Command function to use a ProxyDownstream steve cliente by cluster

### DIFF
--- a/extensions/kubectl/command.go
+++ b/extensions/kubectl/command.go
@@ -96,7 +96,11 @@ func Command(client *rancher.Client, yamlContent *management.ImportClusterYamlIn
 		return "", err
 	}
 
-	steveClient := client.Steve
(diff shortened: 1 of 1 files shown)


extensions/kubectl/command.go
```diff
@@ -96,7 +96,11 @@ func Command(client *rancher.Client, yamlContent *management.ImportClusterYamlIn
 		return "", err
 	}
 
-	steveClient := client.Steve
+	steveClient, err := client.Steve.ProxyDownstream(clusterID)
```
anupama2501, line 99: `steveClient := client.Steve` is needed for local cluster. 

can you add in a condition where if the clusterID is local we set steveClient := client.Steve. if not, steveClient, err := client.Steve.ProxyDownstream(clusterID)

caliskanugur (approved):
LGTM

PR #268, opened 2024-08-22
<v2.8> Fixed pod logs streaming error

lscalabrini01:
Changed kubectl Command function to use a ProxyDownstream steve cliente by cluster

PR #269, opened 2024-08-22
<main> Fixed pod logs streaming error

lscalabrini01:
Changed kubectl Command function to use a ProxyDownstream steve cliente by cluster

Backports:
[v2.9](https://github.com/rancher/shepherd/pull/266)
[v2.8](https://github.com/rancher/shepherd/pull/268)
[v2.7](https://github.com/rancher/shepherd/pull/267)

### DIFF
--- a/extensions/kubectl/command.go
+++ b/extensions/kubectl/command.go
@@ -96,7 +96,11 @@ func Command(client *rancher.Client, yamlContent *management.ImportClusterYamlIn
 		return "", err
 	}
 
-	steveClient := client.Steve
+	steveClient, err := client.Steve.ProxyDownstream(clusterID)
+	if err != nil {
+		return "", err
+	}
+
 	pods, err := steveClient.SteveType(pods.PodResourceSteveType).NamespacedSteveClient(Namespace).List(nil)
 	if err != nil {
 		return "", err


extensions/kubectl/command.go
```diff
@@ -96,7 +96,11 @@ func Command(client *rancher.Client, yamlContent *management.ImportClusterYamlIn
 		return "", err
 	}
 
-	steveClient := client.Steve
+	steveClient, err := client.Steve.ProxyDownstream(clusterID)
```
anupama2501, line 99: steveClient := client.Steve is needed for local cluster.

can you add in a condition where if the clusterID is local we set steveClient := client.Steve. if not, steveClient, err := client.Steve.ProxyDownstream(clusterID)

lscalabrini01: Yes, I can

anupama2501: if this works with local cluster also, then we dont need to add the condition..

lscalabrini01: Actually, it is just working on RKE2 clusters, I tested in RKE1 and K3S and it is not working the if will be necessary

anupama2501 (approved):

slickwarren (approved):

caliskanugur (approved):
Looks good to me

PR #279, opened 2024-08-28
Generic Permutations

susesgartner:
Purpose: 
1. To enable permuting of all fields in the cattle config so we can move forward with OS testing automation
2. To lay some of the groundwork for replacing/deprecating provisioning/permutation

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -119,3 +119,19 @@ func WriteConfig(key string, config interface{}) error {
 
 	return nil
 }
+
+// LoadConfigFromFile loads an entire yaml file into a map[string]any
+func LoadConfigFromFile(filePath string) map[string]any {
+	allString, err := os.ReadFile(filePath)
+	if err != nil {
+		panic(err)
+	}
+
+	var all map[string]any
+	err = yaml.Unmarshal(allString, &all)
+	if err != nil {
+		panic(err)
+	}
+
+	return all
+}
--- a/pkg/config/operations/operations.go
+++ b/pkg/config/operations/operations.go
@@ -1,0 +1,95 @@
+package operations
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"sigs.k8s.io/yaml"
+)
+
+// ReplaceValue recursively traverses keyPath, replacing the specified replaceVal in searchMap.
+func ReplaceValue(keyPath []string, replaceVal any, searchMap map[string]any) (map[string]any, error) {
+	if len(keyPath) <= 1 {
+		searchMap[keyPath[0]] = replaceVal
+
+		return searchMap, nil
+	} else {
+		var err error
+
+		if _, ok := searchMap[keyPath[0]].(map[string]any); ok {
+			searchMap[keyPath[0]], err = ReplaceValue(keyPath[1:], replaceVal, searchMap[keyPath[0]].(map[string]any))
+			if err != nil {
+				return nil, err
+			}
+		} else if _, ok := searchMap[keyPath[0]].([]any); ok {
+			for i := range searchMap[keyPath[0]].([]any) {
+				searchMap[keyPath[0]].([]any)[i], err = ReplaceValue(keyPath[1:], replaceVal, searchMap[keyPath[0]].([]any)[i].(map[string]any))
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	return searchMap, nil
+}
+
+// GetValue recursively traverses keyPath, returning the value of the specified keyPath in searchMap.
+func GetValue(keyPath []string, searchMap map[string]any) (any, error) {
+	var err error
+	var keypathvalues any
+	if len(keyPath) == 1 {
+		keypathvalues, ok := searchMap[keyPath[0]]
+		if !ok {
+			err = errors.New(fmt.Sprintf("expected key does not exist: %s", keyPath[0]))
+		}
+		return keypathvalues, err
+	} else {
+		if _, ok := searchMap[keyPath[0]].(map[string]any); ok {
+			keypathvalues, err = GetValue(keyPath[1:], searchMap[keyPath[0]].(map[string]any))
+			if err != nil {
+				return nil, err
+			}
+		} else if _, ok := searchMap[keyPath[0]].([]any); ok {
+			for i := range searchMap[keyPath[0]].([]any) {
+				keypathvalues, err = GetValue(keyPath[1:], searchMap[keyPath[0]].([]any)[i].(map[string]any))
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+
+	return keypathvalues, err
+}
+
+// LoadObjectFromMap unmarshals a specific key's value into an object
+func LoadObjectFromMap(key string, config map[string]any, object any) {
+	keyConfig := config[key]
+	scopedString, err := yaml.Marshal(keyConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	err = yaml.Unmarshal(scopedString, &object)
+	if err != nil {
+		panic(err)
+	}
+}
+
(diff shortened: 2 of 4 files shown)


extensions/configoperations/configoperations.go
```diff
@@ -0,0 +1,95 @@
...
+		return keypathvalues, err
+	} else {
+		if _, ok := searchMap[keyPath[0]].(map[string]any); ok {
+			keypathvalues, err = GetValue(keyPath[1:], searchMap[keyPath[0]].(map[string]any))
+			if err != nil {
+				return nil, err
+			}
+		} else if _, ok := searchMap[keyPath[0]].([]any); ok {
+			for i := range searchMap[keyPath[0]].([]any) {
+				keypathvalues, err = GetValue(keyPath[1:], searchMap[keyPath[0]].([]any)[i].(map[string]any))
```
slickwarren, line 55: does keypathvalues not need some [i] value here since its in a for loop?

extensions/configoperations/permutations/permutations.go
```diff
@@ -0,0 +1,107 @@
...
+func CreatePermutation(keyPath []string, keyPathValues []any, keyPathValueRelationships []Relationship) Permutation {
+	return Permutation{
+		KeyPath:                   keyPath,
+		KeyPathValues:             keyPathValues,
+		KeyPathValueRelationships: keyPathValueRelationships,
+	}
+}
+
+// Permute iterates over a list of permutation structs and permutes the base config with each of the permutations
+func Permute(permutations []Permutation, baseConfig map[string]any) ([]map[string]any, string, error) {
```
slickwarren, line 44: what is the returned string used for? I only see it set to an empty string in all the returns

slickwarren (commented):
some overarching questions:
* can you provide some example runs i.e. in jenkins of this working?
* can you also provide instructions on how to use this version of permutations

extensions/configoperations/permutations/permutations.go
susesgartner: Your right 😓 I forgot to update that. It is going to return a list of permutation names.

extensions/configoperations/configoperations.go
susesgartner: * I'll get some jenkins runs for this up. I have it hooked up to RKE2DynamicInput in my test branch

* Sure, I could add a readme to Permutations with instructions on how to utilize it. Or would you just prefer a code example in rancher/rancher?

susesgartner: For the keypathvalues in the case that we hit a list of maps I made the choice to have this as the default behavior. 

The reason for this is that if we have a list of maps we run into a scenario where GetValue/ReplaceValue have multiple values to get or replace and would have to choose which ones to perform that operation on. An example of this would be if we wanted to get/replace the ami field and we had multiple machineConfigs, which ami field should we get/replace?

 Currently I don't support that in permutations and frankly there isn't really a scenario where we need to right now.(as far as i know) . If/When there is a reason to update this I will do so however right now only getting/replacing the first value is sufficient.

caliskanugur: Godoc supports examples, in my opinion, the function examples should live where the code lives (maybe not necessarily config-related but how it works and how the output would look). You can also put an example here for the functions, there should be a comment example in v1 clients for converting

slickwarren: what about custom cluster for windows ami? 

```
Or would you just prefer a code example in rancher/rancher?
```
I think in rancher is fine

extensions/configoperations/permutations/permutations.go
susesgartner: After considering further I have decided to remove this functionality from the Permute function. If we want to build the permutation name there are other ways to do so that are more effective.

extensions/configoperations/configoperations.go
susesgartner: Added a godoc example to the permutations package

extensions/configoperations/configoperations.go
```diff
@@ -0,0 +1,111 @@
...
+				}
+			}
+		}
+	}
+
+	return keypathvalues, err
+}
+
+// LoadKeyFromMap unmarshals a specific key's value into a map[string]any
+func LoadKeyFromMap(key string, config map[string]any) map[string]any {
```
slickwarren, line 68: where is this used? the name is a little confusing. But unclear how to make it better without context of its use. Maybe something like `GetChildMapFromKey`?

extensions/configoperations/configoperations.go
```diff
@@ -0,0 +1,111 @@
...
+	err = yaml.Unmarshal(scopedString, &scopedMap)
+	if err != nil {
+		panic(err)
+	}
+
+	return scopedMap
+}
+
+// LoadObjectFromMap unmarshals a specific key's value into an object
+func LoadObjectFromMap(key string, config map[string]any, object any) {
```
slickwarren, line 85: I don't see this used elsewhere either.

pkg/config/config.go
```diff
@@ -119,3 +119,19 @@ func WriteConfig(key string, config interface{}) error {
 
 	return nil
 }
+
+// LoadConfigFromFile loads an entire yaml file into a map[string]any
+func LoadConfigFromFile(filePath string) map[string]any {
```
slickwarren, line 124: is the idea to have a separate file from CATTLE_TEST_CONFIG? 

in other words, why can't we just use the existing 'LoadConfig'?

extensions/configoperations/permutations/permutations.go
```diff
@@ -0,0 +1,103 @@
...
+		KeyPathValueRelationships: keyPathValueRelationships,
+	}
+}
+
+// Permute iterates over a list of permutation structs and permutes the base config with each of the permutations
+func Permute(permutations []Permutation, baseConfig map[string]any) ([]map[string]any, error) {
+	var configs []map[string]any
+	var err error
+	if len(permutations) == 0 {
+		return configs, err
```
slickwarren, line 46: won't you need to throw a new error since both are nil at this step?

pkg/config/config.go
slickwarren: maybe providing a readme of what configs will look like would help?

extensions/configoperations/permutations/example_test.go
```diff
@@ -0,0 +1,37 @@
+package permutations
+
+import (
+	"fmt"
+
+	"github.com/rancher/shepherd/extensions/configoperations"
+)
+
+func ExamplePermute() {
```
markusewalker, line 9: Previously resolved, but can you put a godoc comment? You have examples below, but would be good to have a godoc comment as well.

markusewalker (commented):
Overall, LGTM, left one comment. Will approve once resolved.

extensions/configoperations/permutations/example_test.go
susesgartner: Sorry, looks like I had this change but didn't merge it before. Fixed.

slickwarren (approved):

markusewalker (dismissed):

pkg/config/config.go
caliskanugur: Great question, so, Sam you don't want the second unmarshal happening? Let's leave it this way, later we can update `LoadConfig's` first part with similar to this function.

extensions/configoperations/configoperations.go
caliskanugur, line 1: This looks good, for other `configoperations` files as well; I can't seem to see any relation to tests, or clients. This is a Shepherd-specific feature.  

Let's move them to `pkg/config/operations/operations.go` etc. ?

caliskanugur (commented):
Looks solid, left some questions

caliskanugur (approved):
Looks solid to me, do we need backports for this?

slickwarren (approved):

markusewalker (approved):
LGTM

PR #280, opened 2024-08-29
[2.9] Credentials update

susesgartner:
Convert V3-->V1 objects and also move the loadConfig call outside of the cloud credentials create as it is bad practice to:
1. Hardcode a load config in functional logic that gets reused (reduces reusability)
2. Combine crud operations. (in this case get and create)


igomez06 (dismissed):
LGTM

caliskanugur (dismissed):
LGTM

markusewalker (dismissed):

igomez06 (approved):
LGTM

PR #288, opened 2024-09-04
multiple rancher clients

brudnak:
Purpose:

1. To enable testing more than a single Rancher instance at a time. This is needed to finally automate away the manual checks we do with Hosted/Tenant Rancher. 

Backports:

- 2.9 https://github.com/rancher/shepherd/pull/305
- 2.8 https://github.com/rancher/shepherd/pull/306
- 2.7 https://github.com/rancher/shepherd/pull/307

### DIFF
--- a/clients/rancher/client.go
+++ b/clients/rancher/client.go
@@ -51,23 +51,15 @@ type Client struct {
 	// Session is the session object used by the client to track all the resources being created by the client.
 	Session *session.Session
 	// Flags is the environment flags used by the client to test selectively against a rancher instance.
-	Flags      *environmentflag.EnvironmentFlags
-	restConfig *rest.Config
-	UserID     string
+	Flags            *environmentflag.EnvironmentFlags
+	restConfig       *rest.Config
+	UserID           string
+	RancherInstances []*Client
 }
 
-// NewClient is the constructor to the initializing a rancher Client. It takes a bearer token and session.Session. If bearer token is not provided,
-// the bearer token provided in the configuration file is used.
-func NewClient(bearerToken string, session *session.Session) (*Client, error) {
-	rancherConfig := new(Config)
-	config.LoadConfig(ConfigurationFileKey, rancherConfig)
-
+func initializeClient(rancherConfig *Config, session *session.Session, bearerToken string) (*Client, error) {
 	environmentFlags := environmentflag.NewEnvironmentFlags()
 	environmentflag.LoadEnvironmentFlags(environmentflag.ConfigurationFileKey, environmentFlags)
-
-	if bearerToken == "" {
-		bearerToken = rancherConfig.AdminToken
-	}
 
 	c := &Client{
 		RancherConfig: rancherConfig,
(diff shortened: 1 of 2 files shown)


susesgartner:
LGTM, I really like this idea. I do have one question: for your hosted/tenant testing you won't need the qase reporter to report to two different qase test runs right?

clients/rancher/client.go
```diff
@@ -126,6 +118,48 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
...
+	}
+
+	return initializeClient(rancherConfig, session, bearerToken)
+}
+
+func newAdditionalClient(config Config, session *session.Session) (*Client, error) {
+	return initializeClient(&config, session, config.AdminToken)
+}
+
+func (c *Client) RancherClients() ([]*Client, error) {
```
susesgartner, line 138: nit: could you add a go doc here.

caliskanugur (commented):
Could we discuss this at one of the weekly design meeting?

 I think I am out of context for this case, but is there a reason why it's not happening at the test level but at the client level?

So `NewClient` as it is accepts a `bearerToken`, so theoretically speaking, wouldn't this do the same thing?

a new config:

```
hostTenantTestConfig:
    tenant: bearertoken
    tenantHost: myhost.com
```
    
In the test:

```
hostClient := rancher.NewClient()

LoadAndUpdateConfig(tenantHost - for myhost.com)

tenantClient := rancher.NewClient(hostTenantTestConfig.tenant)
```

Couple things that I am curious about with this design:

- Currently it's per bearer token per client (as it accepts the param and if empty _then_ checks the configuration file), but after the change, only the initial client will have this ability altough there can be multiple clients
- If the recursive configuration is picked, does `NewClient` need a change? Would a loop in `RancherClients` that instantiates `NewClient` for each recursive configuration be enough?


There's also a discussion about decoupling client from the configuration file and allowing accepting the configuration object to instantiate a client, which would eliminate the example's second step too

igomez06:
> Could we discuss this at one of the weekly design meeting?
> 
> I think I am out of context for this case, but is there a reason why it's not happening at the test level but at the client level?
> 
> So `NewClient` as it is accepts a `bearerToken`, so theoretically speaking, wouldn't this do the same thing?
> 
> a new config:
> 
> ```
> hostTenantTestConfig:
>     tenant: bearertoken
>     tenantHost: myhost.com
> ```
> 
> In the test:
> 
> ```
> hostClient := rancher.NewClient()
> 
> LoadAndUpdateConfig(tenantHost - for myhost.com)
> 
> tenantClient := rancher.NewClient(hostTenantTestConfig.tenant)
> ```
> 
> Couple things that I am curious about with this design:
> 
> * Currently it's per bearer token per client (as it accepts the param and if empty _then_ checks the configuration file), but after the change, only the initial client will have this ability altough there can be multiple clients
> * If the recursive configuration is picked, does `NewClient` need a change? Would a loop in `RancherClients` that instantiates `NewClient` for each recursive configuration be enough?
> 
> There's also a discussion about decoupling client from the configuration file and allowing accepting the configuration object to instantiate a client, which would eliminate the example's second step too

clients/rancher/client.go
brudnak: added 🫡 📝

brudnak:
> LGTM, I really like this idea. I do have one question: for your hosted/tenant testing you won't need the qase reporter to report to two different qase test runs right?

No, I won't. A test case checking a host/tenant issue would be reported as a single test scenario, in Qase.

igomez06 (commented):
This looks good, have you tested it Andrew?

brudnak:
> This looks good, have you tested it Andrew?

Thanks @igomez06! Yeah I've tested and posted the results in our internal communication

igomez06 (approved):
LGTM

caliskanugur:
There's a potential conflict with this PR: https://github.com/rancher/shepherd/pull/296

Let's chat about which one goes first in the design meeting.

I think it looks good definitely fixing my recursive comment. My only comment is in the config, just because it's a slice, I want to learn more about how we identify which client is what

PR #296, opened 2024-09-23
Enable creating a rancher client by directly passing a *Config

git-ival:
## Problem
It is not currently possible to create a rancher Client without relying on Shepherd's configuration file load mechanism which generates Config structs as needed.
 
## Solution
Small refactor to the rancher client's `client.go` which pulls out the majority of the client creation logic into a separate function and adds the `NewClientForConfig` function which takes a *Config directly.
 
## Testing
~~Tested locally within the rancher/dartboard project, but have not run any existing tests elsewhere.~~
Ran two rounds of tests both utilizing the new client creation function as well as the default client creation function

There should be no regressions as the flow of the `NewClient` function has not been modified.

Backports:
  - #311 
  - #312
  - #378

### DIFF
--- a/clients/rancher/client.go
+++ b/clients/rancher/client.go
@@ -74,10 +74,32 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 		Flags:         &environmentFlags,
 	}
 
-	session.CleanupEnabled = *rancherConfig.Cleanup
+	return newClient(c, bearerToken, rancherConfig, session)
+}
 
+// NewClientForConfig is the constructor for initializing a rancher Client for the given config and session.
+func NewClientForConfig(bearerToken string, rancherConfig *Config, session *session.Session) (*Client, error) {
+	environmentFlags := environmentflag.NewEnvironmentFlags()
+	environmentflag.LoadEnvironmentFlags(environmentflag.ConfigurationFileKey, environmentFlags)
+
+	if bearerToken == "" {
+		bearerToken = rancherConfig.AdminToken
+	}
+
+	c := &Client{
+		RancherConfig: rancherConfig,
+		Flags:         &environmentFlags,
+	}
+
+	return newClient(c, bearerToken, rancherConfig, session)
+}
+
+func newClient(c *Client, bearerToken string, config *Config, session *session.Session) (*Client, error) {
+	if session != nil {
+		session.CleanupEnabled = *config.Cleanup
+	}
 	var err error
-	restConfig := newRestConfig(bearerToken, rancherConfig)
+	restConfig := newRestConfig(bearerToken, config)
 	c.restConfig = restConfig
 	c.Session = session
 	c.Management, err = management.NewClient(clientOpts(restConfig, c.RancherConfig))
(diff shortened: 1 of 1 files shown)


clients/rancher/client.go
```diff
@@ -68,16 +68,35 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
...
 		RancherConfig: rancherConfig,
 		Flags:         &environmentFlags,
 	}
 
-	session.CleanupEnabled = *rancherConfig.Cleanup
+	return newClient(c, bearerToken, rancherConfig, session)
+}
+
+// NewClientForConfig is the constructor for initializing a rancher Client for the given config and session.
+func NewClientForConfig(bearerToken string, config *Config, session *session.Session) (*Client, error) {
```
anupama2501, line 82: in what case would we want to send in the bearer token separately tho? why not retrieve it from the rancher config itself?

clients/rancher/client.go
```diff
@@ -68,16 +68,35 @@ func NewClient(bearerToken string, session *session.Session) (*Client, error) {
 	if bearerToken == "" {
 		bearerToken = rancherConfig.AdminToken
 	}
+	// rancherConfig.AdminToken = bearerToken
```
anupama2501, line 71: can we remove this comment?

clients/rancher/client.go
git-ival: The bearerToken is used for setting up the restConfig, and the managemnent, v1, wrangler, cli and catalog clients. It allows for things like the `Client.AsUser(user)` function to utilize the rancher client as a non-admin via a non-admin bearertoken that can be passed to the function.

Tests like `tests/v2/integration/projects/project_user_test.go` and actions like `tests/v2/actions/rbac/rbac.go` as well as various other places utilize the `Client.AsUser(user)` function.

anupama2501: got it. In that case do we want to add a check to see if a bearer token is empty and if it is, we can retrieve it from the rancher config itself - similar to how we have it for `NewClient`. will be useful for admin users to not have to send in the token every time?

git-ival: Good call! Adding that in now

anupama2501 (approved):

slickwarren (commented):
I think it would be worth a test run of this in jenkins, otherwise the code lgtm

rancher-max (approved):
LGTM. Waiting on creating dartboard Jenkins jobs for now until we have the new Harvester environment setup so I think this is fine with just your local runs of the new function.

caliskanugur:
Cross-referencing to a great point @rancher-max asked in one of the backports of the PR.

https://github.com/rancher/shepherd/pull/312#pullrequestreview-2609358534

> This backport looks slightly different than the main issue, with where the rancherConfig.AdminToken is being set. Is this intentional? We also might be able to get away with not having a 2.8 backport now, but will need a 2.10 backport still.

What do we think about backports for this? Since it's a client feature, are there any plans to use this feature for other release lines? 

Usually for extensions, we would need backports for testing compatibility, what do you all think?

anupama2501:
I remember Andrew was working on a similar feature and I think it might be worth to add in the release lines if we want to make use of this for any of the features in 2.10, 2.9 or 2.8?

igomez06:
> I remember Andrew was working on a similar feature and I think it might be worth to add in the release lines if we want to make use of this for any of the features in 2.10, 2.9 or 2.8?

Those are my thoughts exactly.

slickwarren (approved):

git-ival:
@caliskanugur All backports should be good to go now, lmk if any further changes are needed 🫡

caliskanugur (approved):
Looks good to me, net improvement, thank you @git-ival!

PR #297, opened 2024-09-24
Add support for setting k8s version in EKS NodeGroup

thehejik:
Adds support for a different k8s version on EKS nodegroup than used on control-plane.

This allows to define an invalid EKS cluster under rancher which is needed for negative testing of Hosted Providers operators.

image

If the nodegroup k8s version is not defined it will use the same version as for control-plane so it should be backward compatible.


igomez06 (commented):
Can you make back  ports? Looks good

thehejik:
> Can you make back ports? Looks good

@igomez06 backports done.

caliskanugur:
Do we need a 2.7 backport as well for this test/feature or are we good with these backports?

igomez06 (approved):
LGTM

thehejik:
> Do we need a 2.7 backport as well for this test/feature or are we good with these backports? @thehejik

@caliskanugur backport PR created, please see above - we are always using `main` module for hosted providers testing even on v2.7, but maybe someone else will use it

PR #304, opened 2024-09-27
improve snapshot creation and validation

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/1525

 
## Problem

snapshots weren't modular enough to directly move to actions, and there was some flaky behavior when using the underlying functions when testing. 
 
## Solution

 * created a new func in clusters.go to wait on snapshots to finish being taken based on observed behavior when taking snapshots
 * improved return of etcd snapshot create function calls to only return a list of snapshots created after the snapshot was initiated
   * since rancher doesn't explicitly do this for you, I added logic to return the created snapshots vs. all snapshots (default rancher behavior) to improve validations later + match behavior of other 'create' functions in shepherd

## Testing

tested on rke1 + rke2 snapshot and restores for most complex automated use case (snapshot, upgrade, restore)

_TODO_
this should be purely improvement to existing functions. Any refactoring of existing use of these functions is taken care of in the rancher/rancher PR.

### DIFF
--- a/extensions/clusters/clusters.go
+++ b/extensions/clusters/clusters.go
@@ -3,7 +3,7 @@ package clusters
 import (
 	"context"
 	"fmt"
-	"strings"
+	"slices"
 	"time"
 
 	"github.com/pkg/errors"
@@ -400,7 +400,12 @@ func WaitClusterToBeInUpgrade(client *rancher.Client, clusterID string) (err err
 	if err != nil {
 		return
 	}
+
 	checkFuncWaitToBeInUpgrade := func(event watch.Event) (bool, error) {
+		acceptableErrorMessages := []string{
+			"Cluster health check failed: Failed to communicate with API server during namespace check",
+			"the object has been modified",
+		}
 		clusterUnstructured := event.Object.(*unstructured.Unstructured)
 		summarizedCluster := summary.Summarize(clusterUnstructured)
 
@@ -408,9 +413,9 @@ func WaitClusterToBeInUpgrade(client *rancher.Client, clusterID string) (err err
 
 		if summarizedCluster.Transitioning && !summarizedCluster.Error && (summarizedCluster.State == clusterStateUpdating || summarizedCluster.State == clusterStateUpgrading) {
 			return true, nil
-		} else if summarizedCluster.Error && isClusterInaccessible(summarizedCluster.Message) {
+		} else if summarizedCluster.Error && isClusterInaccessible(summarizedCluster.Message, acceptableErrorMessages) {
 			return false, nil
-		} else if summarizedCluster.Error && !isClusterInaccessible(summarizedCluster.Message) {
+		} else if summarizedCluster.Error && !isClusterInaccessible(summarizedCluster.Message, acceptableErrorMessages) {
 			return false, errors.Wrap(err, clusterErrorStateMessage)
 		}
 
@@ -440,6 +445,10 @@ func WaitClusterUntilUpgrade(client *rancher.Client, clusterID string) (err erro
 		return
 	}
 	checkFuncWaitUpgrade := func(event watch.Event) (bool, error) {
+		acceptableErrorMessages := []string{
+			"Cluster health check failed: Failed to communicate with API server during namespace check",
+			"the object has been modified",
+		}
 		clusterUnstructured := event.Object.(*unstructured.Unstructured)
 		summarizedCluster := summary.Summarize(clusterUnstructured)
 
@@ -447,9 +456,9 @@ func WaitClusterUntilUpgrade(client *rancher.Client, clusterID string) (err erro
 
 		if summarizedCluster.IsReady() {
 			return true, nil
-		} else if summarizedCluster.Error && isClusterInaccessible(summarizedCluster.Message) {
+		} else if summarizedCluster.Error && isClusterInaccessible(summarizedCluster.Message, acceptableErrorMessages) {
 			return false, nil
-		} else if summarizedCluster.Error && !isClusterInaccessible(summarizedCluster.Message) {
+		} else if summarizedCluster.Error && !isClusterInaccessible(summarizedCluster.Message, acceptableErrorMessages) {
 			return false, errors.Wrap(err, clusterErrorStateMessage)
 
 		}
@@ -483,12 +492,53 @@ func WaitClusterToBeUpgraded(client *rancher.Client, clusterID string) (err erro
 	return
 }
 
-func isClusterInaccessible(messages []string) (isInaccessible bool) {
-	clusterCPErrorMessage := "Cluster health check failed: Failed to communicate with API server during namespace check" // For GKE
-	clusterModifiedErrorMessage := "the object has been modified"                                                        // For provisioning node driver K3s and RKE2
+// WaitOnClusterAfterSnapshot waits for a cluster to finish taking a snapshot and return to an active state.
+func WaitOnClusterAfterSnapshot(client *rancher.Client, clusterID string) error {
+	cluster, err := client.Steve.SteveType(ProvisioningSteveResourceType).ByID(clusterID)
+	if err != nil {
+		return err
+	}
 
+	isTransitioning := cluster.State == nil || cluster.State.Transitioning
+
+	if !isTransitioning {
+		err = kwait.PollUntilContextTimeout(context.TODO(), defaults.FiveHundredMillisecondTimeout, defaults.OneMinuteTimeout, true, func(ctx context.Context) (bool, error) {
+			cluster, err := client.Steve.SteveType(ProvisioningSteveResourceType).ByID(clusterID)
+			if err != nil {
+				return false, err
+			}
+
+			// note, this intentionally ignores cluster.State.Error, as that can sometimes appear during an upgrade during snapshots.
+			if cluster.State == nil {
+				return false, nil
+			}
+			return cluster.State.Transitioning, nil
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	err = kwait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaults.FifteenMinuteTimeout, true, func(ctx context.Context) (bool, error) {
+		cluster, err := client.Steve.SteveType(ProvisioningSteveResourceType).ByID(clusterID)
+		if err != nil {
+			return false, err
+		}
+
+		if cluster.State == nil {
+			return false, nil
+		}
+		// note, this intentionally ignores cluster.State.Error, as that can sometimes appear during an upgrade during snapshots.
+
+		return cluster.State.Name == active, nil
+	})
+
+	return err
+}
+
+func isClusterInaccessible(messages, acceptableErrors []string) (isInaccessible bool) {
 	for _, message := range messages {
-		if strings.Contains(message, clusterCPErrorMessage) || strings.Contains(message, clusterModifiedErrorMessage) {
+		if slices.Contains(acceptableErrors, message) {
 			isInaccessible = true
 			break
 		}
--- a/extensions/etcdsnapshot/etcdsnapshot.go
+++ b/extensions/etcdsnapshot/etcdsnapshot.go
@@ -2,8 +2,9 @@ package etcdsnapshot
 
 import (
 	"context"
-	"sort"
-	"strings"
+	"errors"
+	"fmt"
+	"net/url"
 	"time"
 
 	"github.com/rancher/norman/types"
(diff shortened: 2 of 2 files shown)


extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -2,114 +2,61 @@ package etcdsnapshot
...
-	sort.Slice(snapshots, func(i, j int) bool {
-		return snapshots[i].ObjectMeta.CreationTimestamp.Before(&snapshots[j].ObjectMeta.CreationTimestamp)
-	})
+// CreateRKE1Snapshot is a helper function to create a snapshot on an RKE1 cluster. Returns error if any.
+func CreateRKE1Snapshot(client *rancher.Client, clusterName string) ([]management.EtcdBackup, error) {
 
-	return snapshots, nil
-}
+	updateTimestamp := time.Now().UTC()
+	logrus.Info("UpdateTime: ", updateTimestamp.String())
```
markusewalker, line 39: You can probably just leave this out, but love that all of this can be cut down.

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -127,23 +76,35 @@ func CreateRKE1Snapshot(client *rancher.Client, clusterName string) error {
...
+
+			snapshotTime, err := time.Parse(time.RFC3339, snapshot.Created)
+			if err != nil {
+				return false, err
+			}
+
+			// time.Parse doesn't include nanoseconds, but time.Now() does. Rounding up by 1 Second.
+			snapshotTime = snapshotTime.Add(time.Duration(oneSecond))
+
+			if snapshotTime.Compare(updateTimestamp) > -1 {
```
markusewalker, line 88: Have you tested with multiple snapshots for both local and S3 snapshots? Just to validate this is consistently working in that scenario.

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -258,62 +234,50 @@ func RestoreRKE1Snapshot(client *rancher.Client, clusterName string, snapshotRes
...
-		return false, nil
-	})
+	logrus.Infof("Restoring snapshot: %v", snapshotRestore.Name)
+	updatedCluster, err := client.Steve.SteveType(ProvisioningSteveResouceType).Update(existingSteveAPIObject, steveWithUpdates)
 	if err != nil {
 		return err
 	}
 
-	err = kwait.PollUntilContextTimeout(context.TODO(), 500*time.Millisecond, defaults.FifteenMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
+	err = kwait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaults.TwoMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
```
markusewalker, line 259: For clarification, why was this cut down to two minutes? Context behind the 15 minutes is that at times, RKE2 can take quite a bit of time to fully restore, so that was used as a safeguard to not have false negative tests.

markusewalker (commented):
Left some comments.

extensions/etcdsnapshot/etcdsnapshot.go
slickwarren: this was split into 2 waits; since this part is only waiting on the snapshots to _start_, I cut it down to 2 mins

later (outside of this function) as part of cluster upgrade / cluster validation, the state is checked again where the timeout is longer ~15 mins

I postulate this is more in line with the structure of what the create functions do; the restore has started, which is what we want. Later, we verify that the process(es) are complete in some other validation i.e. in this case its a cluster upgrade check that is done.

extensions/etcdsnapshot/etcdsnapshot.go
slickwarren: yes, tested with local and s3, I'll provide some jenkins jobs in the next day or so.

extensions/etcdsnapshot/etcdsnapshot.go
slickwarren: ack

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -2,114 +2,60 @@ package etcdsnapshot
...
-	"sort"
-	"strings"
+	"errors"
+	"fmt"
+	"net/url"
 	"time"
 
 	"github.com/rancher/norman/types"
 	apisV1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
+	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
```
rancher-max, line 12: Shouldn't need this as it's the same as the package defined above (`apisV1`)

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -2,114 +2,60 @@ package etcdsnapshot
...
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
 	"github.com/rancher/shepherd/clients/rancher"
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
 	rancherv1 "github.com/rancher/shepherd/clients/rancher/v1"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
```
rancher-max, line 17: Shouldn't need this either as it's the same as `rancherv1` defined above

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -2,114 +2,60 @@ package etcdsnapshot
...
+	SnapshotClusterNameLabel     = "rke.cattle.io/cluster-name"
 	fleetNamespace               = "fleet-default"
 	localClusterName             = "local"
 	active                       = "active"
 	readyStatus                  = "Resource is ready"
+	oneSecond                    = 1000000000
 )
 
-// GetRKE1Snapshots is a helper function to get the existing snapshots for a downstream RKE1 cluster.
-func GetRKE1Snapshots(client *rancher.Client, clusterName string) ([]management.EtcdBackup, error) {
```
rancher-max, line 31: Is this GetRKE1Snapshots definitely not used anywhere? Might be dangerous to remove the public function, but it's fine if we can confirm it's not being used at least in rancher/rancher.

rancher-max (commented):
Overall flow and changes LGTM, just a couple of questions/comments.

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -127,23 +75,35 @@ func CreateRKE1Snapshot(client *rancher.Client, clusterName string) error {
...
 				return false, nil
 			}
+
+			snapshotTime, err := time.Parse(time.RFC3339, snapshot.Created)
+			if err != nil {
+				return false, err
+			}
+
+			// time.Parse doesn't include nanoseconds, but time.Now() does. Rounding up by 1 Second.
+			snapshotTime = snapshotTime.Add(time.Duration(oneSecond))
```
lscalabrini01, line 85: Could you use the time.Second constant?

`snapshotTime = snapshotTime.Add(time.Duration(time.Second))`

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -2,114 +2,60 @@ package etcdsnapshot
...
 
 const (
 	ProvisioningSteveResouceType = "provisioning.cattle.io.cluster"
+	SnapshotSteveResourceType    = "rke.cattle.io.etcdsnapshot"
+	SnapshotClusterNameLabel     = "rke.cattle.io/cluster-name"
 	fleetNamespace               = "fleet-default"
 	localClusterName             = "local"
 	active                       = "active"
 	readyStatus                  = "Resource is ready"
+	oneSecond                    = 1000000000
```
lscalabrini01, line 32: If be possible to use the `time.Second` we can remove this line

extensions/etcdsnapshot/etcdsnapshot.go
slickwarren: yes I confirmed; only was used in the tests I moved, and I replaced this functionality

rancher-max (dismissed):

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -127,23 +72,35 @@ func CreateRKE1Snapshot(client *rancher.Client, clusterName string) error {
 			if snapshotObj.State != active {
 				return false, nil
 			}
+
+			snapshotTime, err := time.Parse(time.RFC3339, snapshot.Created)
```
markusewalker, line 76: Missed this before, this is OS agnostic, correct? If I were to run a snapshot test that utilizes this `time.RFC3339` on macOS, Linux, Windows, I will get the same result, right?

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -165,41 +122,58 @@ func CreateRKE2K3SSnapshot(client *rancher.Client, clusterName string) error {
...
 
 	logrus.Infof("Creating snapshot...")
-	_, err = client.Steve.SteveType(clusters.ProvisioningSteveResourceType).Update(clusterSteveObject, clusterObject)
+	updatedCluster, err := client.Steve.SteveType(clusters.ProvisioningSteveResourceType).Update(clusterSteveObject, clusterObject)
 	if err != nil {
-		return err
+		return nil, err
+	}
+
+	updateTimestamp := updatedCluster.ManagedFields[len(updatedCluster.ManagedFields)-1].Time.Time
```
markusewalker, line 130: Calling this out. The linter caught an `ineffassign` that needs to be addressed.

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -165,41 +122,58 @@ func CreateRKE2K3SSnapshot(client *rancher.Client, clusterName string) error {
...
+		query, err := url.ParseQuery(fmt.Sprintf("labelSelector=%s=%s", SnapshotClusterNameLabel, clusterName))
 		if err != nil {
 			return false, nil
 		}
 
-		_, clusterSteveObject, err := clusters.GetProvisioningClusterByName(client, clusterName, fleetNamespace)
+		snapshotSteveObjList, err := client.Steve.SteveType(SnapshotSteveResourceType).List(query)
 		if err != nil {
 			return false, nil
 		}
```
markusewalker, line 149: Can you add a line here for code readability?

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -165,41 +122,58 @@ func CreateRKE2K3SSnapshot(client *rancher.Client, clusterName string) error {
...
+		}
+
+		return true, nil
 	})
-	if err != nil {
-		return err
-	}
 
-	return nil
+	// not registering cleanup func; users do not delete snapshots through rancher
```
markusewalker, line 174: This can probably be removed.

markusewalker (commented):
Left a clarification comment I missed before and calling out an issue the linter called out that needs to be addressed.

extensions/clusters/clusters.go
```diff
@@ -483,12 +492,58 @@ func WaitClusterToBeUpgraded(client *rancher.Client, clusterID string) (err erro
...
+	if err != nil {
+		return err
+	}
+	isTransitioning := false
+
+	if cluster.State == nil {
+		isTransitioning = true
+	} else {
+		isTransitioning = cluster.State.Transitioning
+	}
```
anupama2501, line 507:
```suggestion
isTransitioning := cluster.State == nil || cluster.State.Transitioning
```

extensions/clusters/clusters.go
```diff
@@ -483,12 +492,58 @@ func WaitClusterToBeUpgraded(client *rancher.Client, clusterID string) (err erro
...
+				return false, nil
+			}
+			return cluster.State.Transitioning, nil
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	err = kwait.PollUntilContextTimeout(context.TODO(), 1*time.Second, defaults.FifteenMinuteTimeout, true, func(ctx context.Context) (bool, error) {
```
anupama2501, line 527: Do these states need to happen every time? If thats the case, can we consolidate both the waits? It was a bit harder to follow through and understand why each wait is necessary. If both waits always occur, maybe we could simplify the logic with a single polling something like:

        if cluster.State == nil {
            return false, nil
        }       

        if cluster.State.Transitioning {
            return cluster.State.Name == "active", nil
        }`

extensions/etcdsnapshot/etcdsnapshot.go
anupama2501: why is the snapshotTime required here? is it to get the latest one?

slickwarren: Correct

extensions/etcdsnapshot/etcdsnapshot.go
```diff
@@ -127,23 +72,35 @@ func CreateRKE1Snapshot(client *rancher.Client, clusterName string) error {
...
 	})
-	if err != nil {
-		return err
-	}
 
-	return nil
+	return snapshots, err
 }
 
 // CreateRKE2K3SSnapshot is a helper function to create a snapshot on an RKE2 or k3s cluster. Returns error if any.
```
anupama2501, line 99: can you update the comment and include returns snapshot object also

extensions/etcdsnapshot/etcdsnapshot.go
markusewalker: @anupama2501 yeah, it's for latest snapshot. There has been flakiness in the past with the incorrect snapshot being grabbed. This becomes an issue at the end of the test as resources are deleted and the wrong snapshot causes those resources to not exist.

extensions/etcdsnapshot/etcdsnapshot.go
slickwarren: I actually prefer to keep this in, since it isn't following the same pattern as other functions. is it okay with you to keep the comment?

extensions/clusters/clusters.go
slickwarren: I'm not sure I understand how this helps. 

if !isTransitioning block will wait for the cluster to enter a transitioning state, which we need because we don't want to throw a false positive.
the block outside the if condition is there to wait on the cluster status to return to active.

anupama2501: the reason being to avoid multiple pollings is to simplify it. this is how the above one goes: it first waits for the cluster to enter the transitioning state, similar to the current implementation and once cluster transitioning begins, it continues polling until the cluster gets to the active state.

slickwarren: I think the current statements make enough sense are are doing what we're expecting of the function.

anupama2501 (approved):

markusewalker (approved):

caliskanugur (approved):
Looks good to me

PR #306, opened 2024-10-02
 [BP v2.8] multiple rancher clients

brudnak:
Purpose:

1. To enable testing more than a single Rancher instance at a time. This is needed to finally automate away the manual checks we do with Hosted/Tenant Rancher.


igomez06 (approved):
LGTM

PR #312, opened 2024-10-09
[BP 2.8] Enable creating a rancher client by directly passing a *Config

git-ival:
## Issue: 
See #296


rancher-max (commented):
This backport looks slightly different than the main issue, with where the rancherConfig.AdminToken is being set. Is this intentional? We also might be able to get away with not having a 2.8 backport now, but will need a 2.10 backport still.

git-ival:
Not intentional, this backport hasn't been updated yet since it was created months ago and was waiting on various test runs + reviews. Will update ASAP

caliskanugur (approved):
LGTM

PR #330, opened 2024-10-25
[2.8] Add cluster.cattle.io, RBAC and Batch Controllers, Add Missing Import and Session Replaces to the Missing Controllers

caliskanugur:
## Problem

RBAC, `cluster.cattle.io ` and batch controllers were missing, as well as some controllers' replaces for session and imports
 
## Solution

RBAC, `cluster.cattle.io ` and batch controllers, add missing map items for the path walks


igomez06 (approved):
LGTM

PR #335, opened 2024-10-30
Update rancher/pkg/apis and use BackingNamespace for projects

JonCrowther:
## Problem
The dependency `rancher/pkg/apis` has changed a fair bit, but recently the linked PR introduced a new field on projects `BackingNamespace` that is used to identify the namespace PRTBs exist in. 
 
## Solution
Updated the dependency and added logic to use the new project field.
 
_TODO_


caliskanugur (approved):
LGTM

PR #342, opened 2024-11-07
adding client for harvester

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/468
 
## Problem

currently, we don't have a way to write automation for harvester using the framework. 
 
## Solution

harvester's API is based on rancher's steve. In this PR I've introduced a harvester client that wraps steve and catalog clients, introduced a new harvesterConfig, similar to how rancherClient is setup. 

We played around with wrapping Steve's API/client directly, but since harvester will need the catalog client as well, giving harvester its own client makes more sense in the long run + to support any use cases their team may have. 

Also played around with simplifying the copy-paste logic between rancher and now harvester clients. But this may make more sense to do if we have to introduce another client that uses most of rancher's API which I don't know of a product that would need it tbh. If this approach were taken, the code would be refactored instead of net-new, risking stuff breaking. 
 
## Testing

see linked rancher PR; we are now able to use harvester's API and import an existing harvester cluster into an existing rancher setup. 

_TODO_

this is a new client, so no risk of regression.

### DIFF
--- a/clients/harvester/client.go
+++ b/clients/harvester/client.go
@@ -16,7 +16,6 @@ import (
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 
 	kubeProvisioning "github.com/rancher/shepherd/clients/provisioning"
-	kubeRKE "github.com/rancher/shepherd/clients/rke"
 	"github.com/rancher/shepherd/pkg/clientbase"
 	"github.com/rancher/shepherd/pkg/config"
 	"github.com/rancher/shepherd/pkg/environmentflag"
@@ -26,10 +25,9 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 )
 
-// Client is the main harvester Client object that gives an end user access to the Provisioning and Management
+// Client is the main harvester Client object that gives an end user access to the Provisioning and Catalog
 // clients in order to create resources on harvester
 type Client struct {
 	// Client used to access Steve v1 API resources
@@ -180,7 +178,7 @@ func (c *Client) GetClusterCatalogClient(clusterID string) (*catalog.Client, err
 }
 
 // GetRancherDynamicClient is a helper function that instantiates a dynamic client to communicate with the harvester host.
-func (c *Client) GetRancherDynamicClient() (dynamic.Interface, error) {
+func (c *Client) GetHarvesterDynamicClient() (dynamic.Interface, error) {
 	dynamic, err := frameworkDynamic.NewForConfig(c.Session, c.restConfig)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


clients/harvester/config.go
```diff
@@ -0,0 +1,13 @@
+package harvester
+
+// The json/yaml config key for the harvester config
+const ConfigurationFileKey = "harvester"
+
+// Config is configuration need to test against a harvester instance
+type Config struct {
```
anupama2501, line 7: this is the same as rancher [Config](https://github.com/rancher/shepherd/blob/main/clients/rancher/config.go#L7), why not re use the existing one by adding a flag or something as such?

anupama2501 (commented):
Main question:

Since all these clients and configurations are already defined within the Rancher [client](https://github.com/rancher/shepherd/blob/main/clients/rancher/client.go#L38), I wonder if we need to have a separate Harvester client. It might be helpful to have a broader discussion to see if a new client is needed or if we might be duplicating existing functionality.

clients/harvester/config.go
slickwarren: had an offline conversation to help with this understanding. lmk if you have any more questions.

clients/harvester/client.go
```diff
@@ -0,0 +1,347 @@
...
+	catalogClient, err := catalog.NewForConfig(&restConfig, c.Session)
+	if err != nil {
+		return nil, err
+	}
+
+	return catalogClient, nil
+}
+
+// GetRancherDynamicClient is a helper function that instantiates a dynamic client to communicate with the harvester host.
+func (c *Client) GetRancherDynamicClient() (dynamic.Interface, error) {
```
anupama2501, line 183: can we rename it to GetHarvesterDynamicClient?

clients/harvester/client.go
```diff
@@ -0,0 +1,347 @@
...
+	provClient, err := kubeProvisioning.NewForConfig(c.restConfig, c.Session)
+	if err != nil {
+		return nil, err
+	}
+
+	return provClient, nil
+}
+
+// GetKubeAPIRKEClient is a function that instantiates a rke client that communicates with the Kube API of a cluster
+func (c *Client) GetKubeAPIRKEClient() (*kubeRKE.Client, error) {
```
anupama2501, line 202: have a question on this as well, since the clusters will be created on rancher itself, do we need to have this function for harvester client? because we create clusters on rancher itself so we can make use of rancher client isn't it?

anupama2501 (commented):
Have a few more questions and comments

clients/harvester/client.go
```diff
@@ -0,0 +1,347 @@
...
+		return nil, err
+	}
+
+	return dynamic, nil
+}
+
+// GetManagementWatchInterface is a functions used to get a watch.Interface from a resource created by the Management Client.
+// As is the Management resources do not have a watch.Interface, so therefore, the dynamic Client is used to get the watch.Interface.
+// The `schemaType` is a string that is found in different Management clients packages. Ex) management.ProjectType
+func (c *Client) GetManagementWatchInterface(schemaType string, opts metav1.ListOptions) (watch.Interface, error) {
```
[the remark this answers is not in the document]
slickwarren: It will still be convenient to have access to management, through V1 instead of V3.

PR #348, opened 2024-11-25
[Main] Fix corral file wait and update the debugging option

caliskanugur:
## Problem

Corral file wait should account for the nil return value from `os.Stat`, which is the state if the file doesn't exist. Currently, the separated command solution makes recurring runs fail because it doesn't wait for the file.
 
## Solution

Account nil return value from `os.Stat`, update to debug flag which gives more output for corral creation.

### DIFF
--- a/clients/corral/corral.go
+++ b/clients/corral/corral.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	debugFlag               = "--trace"
+	debugFlag               = "--debug"
 	skipCleanupFlag         = "--skip-cleanup"
 	corralPrivateSSHKey     = "corral_private_key"
 	corralPublicSSHKey      = "corral_public_key"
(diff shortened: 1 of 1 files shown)


slickwarren:
in the linked issues, its mentioned `and should default to true rather than false`. Will that be done in a separate PR or is there room to do it here?

caliskanugur:
Nice catch, removing that part from the ticket. That part of the ticket was for Airgap defaults. Which is in JJB. 

In recurring runs, it already [defaults to true.](https://github.com/rancher/rancher/blob/main/tests/v2/validation/pipeline/rancherha/main.go#L43)

Also linked the second issue - in PR's description.

slickwarren (approved):
code looks good. Do you have a jenkins job to show this fixes the nil pointer issue?

caliskanugur:
Discussed the sanity checks offline

markusewalker (approved):

anupama2501 (approved):

bmdepesa (approved):

PR #351, opened 2024-11-25
Add harvester Client

slickwarren:
## Issue: 

https://github.com/rancher/qa-tasks/issues/468
 
## Problem

currently, we don't have a way to write automation for harvester using the framework. 
 
## Solution

harvester's API is based on rancher's steve. In this PR I've introduced a harvester client that wraps steve and catalog clients, introduced a new harvesterConfig, similar to how rancherClient is setup. 

We played around with wrapping Steve's API/client directly, but since harvester will need the catalog client as well, giving harvester its own client makes more sense in the long run + to support any use cases their team may have. 

Also played around with simplifying the copy-paste logic between rancher and now harvester clients. But this may make more sense to do if we have to introduce another client that uses most of rancher's API which I don't know of a product that would need it tbh. If this approach were taken, the code would be refactored instead of net-new, risking stuff breaking. 
 
## Testing

see linked rancher PR; we are now able to use harvester's API and import an existing harvester cluster into an existing rancher setup. 

_TODO_

this is a new client, so no risk of regression.

### DIFF
--- a/clients/harvester/client.go
+++ b/clients/harvester/client.go
@@ -1,0 +1,101 @@
+package harvester
+
+import (
+	"fmt"
+
+	frameworkDynamic "github.com/rancher/shepherd/clients/dynamic"
+	"github.com/rancher/shepherd/clients/rancher/catalog"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
+
+	"github.com/rancher/shepherd/pkg/clientbase"
+	"github.com/rancher/shepherd/pkg/config"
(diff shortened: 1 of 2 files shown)


lscalabrini01 (approved):
LGTM

anupama2501 (approved):

floatingman (approved):
LGTM

caliskanugur (approved):
LGTM, do we need backports for this PR or just main?

slickwarren:
yes we will need backports, I wanted to wait for your review to open them 👍🏼 I'll do that now.

PR #356, opened 2024-12-06
Updating the state file for rke1 clusters

anupama2501:
## Issue: 
https://github.com/rancher/qa-tasks/issues/1295
 
## Problem
RKE1 cluster upgrade fails as the state file stays in a secret instead of a config map. This has been previously in a configmap.

## Solution
Updates the state file used in the code from a config map to a state map.
 
## Testing
Jenkins log

_TODO_

### DIFF
--- a/clients/rkecli/state.go
+++ b/clients/rkecli/state.go
@@ -12,7 +12,6 @@ import (
 	"github.com/rancher/shepherd/clients/rancher"
 	v3 "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
-	"github.com/rancher/shepherd/extensions/configmaps"
 	"github.com/rancher/shepherd/pkg/config"
 	"github.com/rancher/shepherd/pkg/file"
 	"gopkg.in/yaml.v2"
@@ -129,7 +128,7 @@ func NewClusterFile(state *cluster.FullState, dirName string, config *Config) (c
 			rkeConfig.Nodes[i].SSHKeyPath = appendSSHPath(rkeConfig.Nodes[i].SSHKeyPath, config.SSHPath)
 		}
 	} else {
-		return "", errors.Wrap(err, "rke SSHPath or SSHKey not configured")
+		return "", errors.New("missing SSH configuration")
 	}
 
 	marshaled, err := yaml.Marshal(rkeConfig)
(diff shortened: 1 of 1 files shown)


lscalabrini01 (dismissed):
LGTM

floatingman (approved):
LGTM

clients/rkecli/state.go
```diff
@@ -129,7 +128,7 @@ func NewClusterFile(state *cluster.FullState, dirName string, config *Config) (c
 			rkeConfig.Nodes[i].SSHKeyPath = appendSSHPath(rkeConfig.Nodes[i].SSHKeyPath, config.SSHPath)
 		}
 	} else {
-		return "", errors.Wrap(err, "rke SSHPath or SSHKey not configured")
+		return "", errors.Wrap(errors.New("missing SSH configuration"), "rke SSHPath or SSHKey not configured")
```
slickwarren, line 131: could you change the func definition to be an unnamed error instead of err. Since it isn't being set, its confusing to keep it in.

clients/rkecli/state.go
```diff
@@ -166,28 +165,27 @@ func NewStateFile(state *cluster.FullState, dirName string) (stateFilePath strin
...
 
-// GetFullState is a function that gets RKE full state from "full-cluster-state" configmap.
+// GetFullState is a function that gets RKE full state from "full-cluster-state" secret.
 // And returns the cluster full state.
 func GetFullState(client *rancher.Client) (state *cluster.FullState, err error) {
-	namespacedConfigmapClient := client.Steve.SteveType(configmaps.ConfigMapSteveType).NamespacedSteveClient(cluster.SystemNamespace)
-	if err != nil {
-		return
-	}
+	namespacedSecretClient := client.Steve.SteveType("secret").NamespacedSteveClient(cluster.SystemNamespace)
```
slickwarren, line 171: was configmap used for any version of 2.8 that we may need to not backport this to?

slickwarren (commented):
some nits / clarifying questions

clients/rkecli/state.go
anupama2501: err is being set in the below line https://github.com/rancher/shepherd/blob/main/clients/rkecli/state.go#L142

clients/rkecli/state.go
anupama2501: This change was introduced in 2.8 versions of rancher and it needs to be back ported to release/v2.8.

caliskanugur: @markusewalker is this related to what we discussed offline back then?

markusewalker: correct, replied offline.

markusewalker (approved):

clients/rkecli/state.go
caliskanugur, line 131: What do you think about removing the wrap since the first part is always nil? Something like this?

```suggestion
		return "", errors.New("missing SSH configuration")
```

caliskanugur (commented):
Looks good to me, just a question, also do we need backports for this, or only 2.8 missing for this change?

lscalabrini01 (approved):
LGTM

caliskanugur (approved):
LGTM

floatingman (approved):

lscalabrini01 (approved):

PR #360, opened 2024-12-19
Fix generic permutation relationships.

susesgartner:
## Problem
Found a bug with generic permutations that caused relationships to not be applied correctly when multiple relationships were present.

Note: since this is limited to relationships, the main use case of generic permutations should be unaffected.
 
## Solution
Fix relationships to appropriately apply when multiple relationships are present and move the logic into its own function for clarity.
 
## Testing
Tested locally as part of the os check development.

### DIFF
--- a/pkg/config/operations/permutations/permutations.go
+++ b/pkg/config/operations/permutations/permutations.go
@@ -9,8 +9,6 @@ import (
 // Relationship structs are used to create an association between a parent value and either a set of permutations or the value of a different key
 type Relationship struct {
 	ParentValue       any           `json:"parentValue" yaml:"parentValue"`
-	ChildKeyPath      []string      `json:"childKeyPath" yaml:"childkeyPath"`
-	ChildKeyPathValue any           `json:"childKeyPathValue" yaml:"childkeyPathValue"`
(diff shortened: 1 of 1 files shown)


pkg/config/operations/permutations/permutations.go
```diff
@@ -104,3 +81,45 @@ func Permute(permutations []Permutation, baseConfig map[string]any) ([]map[strin
 
 	return finalConfigs, err
 }
+
+// applyRelationships iterates over a list of relationships structs and applies them to a config
+func applyRelationships(config map[string]any, relationships []Relationship, keyPathValue any) ([]map[string]any, error) {
```
slickwarren, line 86: could you leave comments for the logic here and look into the deep copies a little more?

susesgartner: Sure. I'll look into the deep copies a bit more.

markusewalker (approved):
LGTM, will need backports.

slickwarren (approved):

caliskanugur (commented):
LGTM, do we need backports for this?

caliskanugur (approved):

Josh-Diamond (approved):

PR #365, opened 2025-01-13
Define hosted providers NodePools and NodeGroups schema as a pointer

valaparthvi:
## Issue: 

Some API requests sent via r/shepherd do not yield the same result as those sent directly to r/rancher, even though the Specs are similar for both, especially when setting a key to an empty object or nil.

## Problem

While automating a test case for EKS to test if deleting all nodegroups resulted in error, I discovered that sending the request via r/shepherd did not give the same result as when sending a request directly to r/rancher. 

The following happens when a request is sent:
>- raw JSON request -> server: server catches the empty array/slice, therefore safety net error thrown
>- shepherd -> JSON encoding (it loses the omitempty fields, not just this one but others omitempties too) -> request -> server: cause it was not there at the first place, it’s lost even before we made the request, no errors cause no changes related to that in that request body
 
## Solution

>Shepherd objects should look more alike raw JSON requests (via network tab on browsers, and unless if there’s a specific reason not to do so) rather than the server side to validate the behavior.

Modify the specs in r/shepherd to look more like raw JSON requests, in this case, let NodePools and NodeGroups be a pointer.

_TODO_


clients/rancher/generated/management/v3/zz_generated_cluster_role_template_binding_status.go
```diff
@@ -0,0 +1,24 @@
+package client
```
valaparthvi, line 1: I am not sure why this file was generated when I ran `go generate`, but I have added it for now.

caliskanugur: Discussed this one offline

caliskanugur (commented):
Looks good to me, solid fix! Do we need backports for this or are we good with the main changes?

caliskanugur (approved):

valaparthvi:
@caliskanugur Does that mean we can merge the changes? I am still running tests against these changes in our repo, but overall they look good.
GKE: https://github.com/rancher/hosted-providers-e2e/actions/runs/12933035721/job/36070731885
EKS: https://github.com/rancher/hosted-providers-e2e/actions/runs/12932695849/job/36069582909
AKS: https://github.com/rancher/hosted-providers-e2e/actions/runs/12925283850

valaparthvi:
>Do we need backports for this or are we good with the main changes?

It's fine, we don't need backports.

PR #373, opened 2025-02-10
Create steve helpers for OS checks

susesgartner:
Related Rancher PR: https://github.com/rancher/tests/pull/37
Description: 
1. Created a new func for accessing the ec2 client which will be used for accessing ami data for the OS checks
2. Made some changes to the steve.go to fix a bug and to allow it to be more generic so it can be used in a node reboot helper. (this includes updates to all functions that utilze this)
3. Add a few constants for 1. and 2.

### DIFF
--- a/clients/ec2/client.go
+++ b/clients/ec2/client.go
@@ -5,6 +5,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/rancher/shepherd/extensions/cloudcredentials"
 	"github.com/rancher/shepherd/pkg/config"
 )
 
@@ -30,10 +31,31 @@ func NewClient() (*Client, error) {
 		return nil, err
 	}
 
-	// Create EC2 service client
 	svc := ec2.New(sess)
 	return &Client{
 		SVC:          svc,
 		ClientConfig: awsEC2ClientConfig,
+	}, nil
+}
+
+// NewClientFromConfig generates a new ec2 client using the provided credentials
+func NewClientFromConfig(awsCredentials *cloudcredentials.AmazonEC2CredentialConfig) (*Client, error) {
+	credential := credentials.NewStaticCredentials(awsCredentials.AccessKey, awsCredentials.SecretKey, "")
+	ec2Session, err := session.NewSession(&aws.Config{
+		Credentials: credential,
+		Region:      aws.String(awsCredentials.DefaultRegion)},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	svc := ec2.New(ec2Session)
+	return &Client{
+		SVC: svc,
+		ClientConfig: &AWSEC2Configs{
+			AWSAccessKeyID:     awsCredentials.AccessKey,
+			AWSSecretAccessKey: awsCredentials.SecretKey,
+			Region:             awsCredentials.DefaultRegion,
+		},
 	}, nil
 }
--- a/extensions/cloudcredentials/aws/create.go
+++ b/extensions/cloudcredentials/aws/create.go
@@ -7,11 +7,16 @@ import (
 	"github.com/rancher/shepherd/extensions/defaults"
 	"github.com/rancher/shepherd/extensions/defaults/namespaces"
 	"github.com/rancher/shepherd/extensions/defaults/providers"
+	"github.com/rancher/shepherd/extensions/defaults/stevestates"
 	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
 	"github.com/rancher/shepherd/extensions/steve"
 	"github.com/rancher/shepherd/pkg/namegenerator"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	localCluster = "local"
 )
 
 // CreateAWSCloudCredentials is a helper function that creates V1 cloud credentials and waits for them to become active.
@@ -35,7 +40,7 @@ func CreateAWSCloudCredentials(client *rancher.Client, credentials cloudcredenti
 		Type: corev1.SecretTypeOpaque,
 	}
 
-	awsCloudCredentials, err := steve.CreateAndWaitForResource(client, stevetypes.Secret, spec, true, defaults.FiveSecondTimeout, defaults.FiveMinuteTimeout)
+	awsCloudCredentials, err := steve.CreateAndWaitForResource(client, namespaces.FleetLocal+"/"+localCluster, stevetypes.Secret, spec, stevestates.Active, defaults.FiveSecondTimeout, defaults.FiveMinuteTimeout)
 	if err != nil {
 		return nil, err
 	}
--- a/extensions/cloudcredentials/aws_config.go
+++ b/extensions/cloudcredentials/aws_config.go
@@ -7,5 +7,5 @@ const AmazonEC2CredentialConfigurationFileKey = "awsCredentials"
 type AmazonEC2CredentialConfig struct {
 	AccessKey     string `json:"accessKey" yaml:"accessKey"`
 	SecretKey     string `json:"secretKey" yaml:"secretKey"`
-	DefaultRegion string `json:"defaultRegion,omitempty" yaml:"defaultRegion"`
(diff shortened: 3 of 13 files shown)


extensions/steve/steve.go
```diff
@@ -63,20 +62,53 @@ func WaitForSteveResourceCreation(client *rancher.Client, v1Resource *v1.SteveAP
 	return err
 }
 
-// CreateResource creates a steve resource and polls the resulting object until it comes up active.
-func CreateAndWaitForResource(client *rancher.Client, v1ResourceType string, v1Resource any, poll bool, interval, timeout time.Duration) (*v1.SteveAPIObject, error) {
-	resource, err := client.Steve.SteveType(v1ResourceType).Create(v1Resource)
+// CreateAndWaitForResource creates a steve resource and polls the resulting object until it is in the desired state.
+func CreateAndWaitForResource(client *rancher.Client, clusterID, v1ResourceType string, v1Resource any, state string, interval, timeout time.Duration) (*v1.SteveAPIObject, error) {
```
bmdepesa, line 66: 1) Modifying this method will break backwards compatibility for other teams/tests using Shepherd -- is there a specific reason we're changing this rather than adding a new method targeting Steve specifically? 

2) Nit - `state` -> `desiredState` or `expectedState`(but this is secondary to the design question)

susesgartner: 1.  I am finishing up the rancher side PR right now and will link it to this one shortly. 
2. I can update that to be a bit more clear.

extensions/steve/steve.go
```diff
@@ -5,24 +5,27 @@ import (
...
-	notFound = "not found"
-	active   = "active"
+	notFound     = "not found"
+	localCluster = "local"
 )
 
-// WaitForSteveResourceDeletion accepts a client, steve resource type, and resource ID, then waits for a steve resource to be deleted
-func WaitForSteveResourceDeletion(client *rancher.Client, interval, timeout time.Duration, steveResourceType, steveID string) error {
+// WaitForResourceDeletion waits for a given steve object to be deleted.
+func WaitForResourceDeletion(client *v1.Client, v1Resource *v1.SteveAPIObject, interval, timeout time.Duration) error {
```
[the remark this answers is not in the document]
susesgartner: From my understanding the only test utilizing this function is the delete_init test in rancher/rancher which I updated as part of the rancher PR associated with this PR.

extensions/cloudcredentials/aws/create.go
```diff
@@ -7,13 +7,18 @@ import (
...
+	"github.com/rancher/shepherd/extensions/defaults/stevestates"
 	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
 	"github.com/rancher/shepherd/extensions/steve"
 	"github.com/rancher/shepherd/pkg/namegenerator"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	localCluster = "local"
```
anupama2501, line 19: can we move localCLuster to defaults as well as its used in all of the cloud providers.

extensions/steve/steve.go
```diff
@@ -63,20 +62,53 @@ func WaitForSteveResourceCreation(client *rancher.Client, v1Resource *v1.SteveAP
...
 		if err != nil {
 			return resource, err
 		}
 	}
 
 	return resource, nil
 }
+
+// GetDownstreamClusterClient fetches the the client of a downstream cluster
+func GetDownstreamClusterClient(client *rancher.Client, clusterID string) (*v1.Client, error) {
```
anupama2501, line 90: since we are creating a client for both downstream cluster as well as local cluster, can the function name be changed to something generic like GetClusterClient()?

extensions/steve/steve.go
```diff
@@ -63,20 +62,53 @@ func WaitForSteveResourceCreation(client *rancher.Client, v1Resource *v1.SteveAP
...
+	var downstreamClusterClient *v1.Client
+	var err error
+	if clusterID != localCluster {
+		cluster, err := client.Steve.SteveType(stevetypes.Provisioning).ByID(clusterID)
+		if err != nil {
+			return nil, err
+		}
+
+		status := &provv1.ClusterStatus{}
```
anupama2501, line 99: why do we want to check the status of the cluster tho here?

extensions/cloudcredentials/aws_config.go
```diff
@@ -5,7 +5,7 @@ const AmazonEC2CredentialConfigurationFileKey = "awsCredentials"
 
 // AmazonEC2CredentialConfig is configuration need to create an aws cloud credential
 type AmazonEC2CredentialConfig struct {
-	AccessKey     string `json:"accessKey" yaml:"accessKey"`
-	SecretKey     string `json:"secretKey" yaml:"secretKey"`
-	DefaultRegion string `json:"defaultRegion,omitempty" yaml:"defaultRegion"`
+	AccessKey     string `json:"accessKey,omitempty" yaml:"accessKey,omitempty"`
```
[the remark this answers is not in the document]
susesgartner: Good point.

extensions/cloudcredentials/aws/create.go
susesgartner: I didn't move this to the defaults package since it doesn't really have a category it fits into, i.e if I put it in the defaults package it would be a single const in a file.

markusewalker: I agree with Sam, if it's going to just be a single const, I'd prefer it stays here.

extensions/steve/steve.go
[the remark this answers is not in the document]
markusewalker: For further context, this function was specifically made for the `delete_init_test.go` test. The idea back then was to have a generalized function that can wait for Steve-based objects.

The reality is that it was never utilized past the `delete_init_test.go` test. If there are plans to use this function for other tests, then I would agree to keep it. If not, I don't see harm in the refactor.

extensions/steve/steve.go
susesgartner: Good point 😓 I honestly don't know why I did it that way at the time. I'll double check this.

clients/ec2/client.go
```diff
@@ -37,3 +38,24 @@ func NewClient() (*Client, error) {
 		ClientConfig: awsEC2ClientConfig,
 	}, nil
 }
+
+func NewClientFromConfig(awsCredentials *cloudcredentials.AmazonEC2CredentialConfig) (*Client, error) {
```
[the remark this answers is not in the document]
susesgartner: Good point.

extensions/steve/steve.go
susesgartner: Took a second look at this and the reason for this is because we need the status.ClusterName to get the proxyDownstream. If you have a better way to get the proxyDownstream I am fine with utilizing that however this currently follows other examples of getting the proxyDownstream that we have in shepherd.

markusewalker (dismissed):

Josh-Diamond (approved):
🚀

extensions/steve/steve.go
susesgartner: The short answer to 1 is that yes it will break backwards compatibility however we wrote the function and I am not aware of any uses of it outside of ours. Adding new methods is fine however over time this kind of practice creates bloat so I try to avoid duplicating code whenever possible since it leads to code bloat.

extensions/steve/steve.go
caliskanugur: Isn't `clusterID == steve cluster name` here as the param?

I've got a few questions here;

1. are these two equivalent? 

```
clusterClient, err = client.Steve.ProxyDownstream(status.ClusterName) & clusterClient, err = client.Steve.ProxyDownstream(clusterID) 
```
2. If I remember correctly, the `clusterID` is local for the local cluster. Does it need to have a condition on this? Would this be enough both for local and downstream clusters?

`clusterClient, err = client.Steve.ProxyDownstream(clusterID)`

susesgartner: Updated to cut out the conditional.

caliskanugur (approved):
Client bits LGTM, do we need backports for the changes?

susesgartner:
We will i'll get those up sometime today.

markusewalker (dismissed):

slickwarren (approved):

anupama2501 (approved):

markusewalker (approved):

Josh-Diamond (approved):

PR #384, opened 2025-03-11
adding extension to get SA'd kubeconfig from harvester via rancher client

slickwarren:
## Issue: 

 https://github.com/rancher/tests/pull/43

Rancher UI specifys a serviceAccountName when generating a kubeconfig from the imported harvester cluster within rancher, using the k8s/clusters api. We need a helper to do this using the apiBaseClient from rancher.

### DIFF
--- a/extensions/harvester/sakubeconfig.go
+++ b/extensions/harvester/sakubeconfig.go
@@ -3,6 +3,7 @@ package harvester
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -31,11 +32,6 @@ type KubeConfigOpts struct {
 // cluster which will use the kubeconfig. This is typically used when creating a downstream cluster for
 // harvester that uses the harvester cloud provider.
 func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byte, error) {
-	client, err := rancher.NewClient(client.RancherConfig.AdminToken, client.Session)
-	if err != nil {
-		return nil, err
-	}
-
 	query, err := url.ParseQuery(HarvesterProviderClusterLabel)
 	if err != nil {
 		return nil, err
@@ -44,6 +40,10 @@ func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byt
 	harvesterCluster, err := client.Steve.SteveType(clusters.ProvisioningSteveResourceType).ListAll(query)
 	if err != nil {
 		return nil, err
+	}
+
+	if len(harvesterCluster.Data) == 0 {
+		return nil, errors.New("no imported harvester cluster found")
 	}
 
 	status := &provv1.ClusterStatus{}
--- a/extensions/harvester/sakubeconfig.go
+++ b/extensions/harvester/sakubeconfig.go
@@ -48,6 +48,9 @@ func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byt
 
 	status := &provv1.ClusterStatus{}
 	err = steveV1.ConvertToK8sType(harvesterCluster.Data[0].Status, status)
+	if err != nil {
+		return nil, err
+	}
 
 	opts := KubeConfigOpts{
 		CSIClusterRoleName: "harvesterhci.io:csi-driver",
--- a/extensions/harvester/sakubeconfig.go
+++ b/extensions/harvester/sakubeconfig.go
@@ -69,7 +69,7 @@ func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byt
 		return nil, err
 	}
 
-	req.Header.Add("Authorization", "Bearer "+client.RancherConfig.AdminToken)
+	req.Header.Add("Authorization", "Bearer "+client.Management.APIBaseClient.Opts.TokenKey)
 	req.Header.Set("Content-Type", "application/json")
 
 	response, err := client.Management.APIBaseClient.Ops.Client.Do(req)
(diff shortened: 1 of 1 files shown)


extensions/harvester/sakubeconfig.go
```diff
@@ -0,0 +1,93 @@
...
+	ClusterRoleName    string `yaml:"clusterRoleName" json:"clusterRoleName" default:"harvesterhci.io:cloudprovider"`
+	Namespace          string `yaml:"namespace" json:"namespace" default:"default"`
+	SaName             string `yaml:"serviceAccountName" json:"serviceAccountName" default:"rancherCluster"`
+}
+
+// GetHarvesterSAKubeconfig generates a kubeconfig from the harvester cluster with a custom SA for the
+// cluster which will use the kubeconfig. This is typically used when creating a downstream cluster for
+// harvester that uses the harvester cloud provider.
+func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byte, error) {
+	client, err := rancher.NewClient(client.RancherConfig.AdminToken, client.Session)
```
igomez06, line 34: can you rename this admin client? So we can see what calls it's making

extensions/harvester/sakubeconfig.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	query, err := url.ParseQuery(HarvesterProviderClusterLabel)
+	if err != nil {
+		return nil, err
+	}
+
+	harvesterCluster, err := client.Steve.SteveType(clusters.ProvisioningSteveResourceType).ListAll(query)
```
igomez06, line 44: maybe check for the list of harvester clusters, so we don't get a panic. Empty list check

igomez06 (commented):
some small nits

extensions/harvester/sakubeconfig.go
slickwarren: added a check after checking err

extensions/harvester/sakubeconfig.go
```diff
@@ -0,0 +1,98 @@
...
+	CSIClusterRoleName string `yaml:"csiclusterRoleName" json:"csiclusterRoleName" default:"harvesterhci.io:csi-driver"`
+	ClusterRoleName    string `yaml:"clusterRoleName" json:"clusterRoleName" default:"harvesterhci.io:cloudprovider"`
+	Namespace          string `yaml:"namespace" json:"namespace" default:"default"`
+	SaName             string `yaml:"serviceAccountName" json:"serviceAccountName" default:"rancherCluster"`
+}
+
+// GetHarvesterSAKubeconfig generates a kubeconfig from the harvester cluster with a custom SA for the
+// cluster which will use the kubeconfig. This is typically used when creating a downstream cluster for
+// harvester that uses the harvester cloud provider.
+func GetHarvesterSAKubeconfig(client *rancher.Client, clusterName string) ([]byte, error) {
```
anupama2501, line 34: do we need the client for any rbac related cases like a standard user for instance? if not we can just take in adminClient as the client is not used anywhere other than getting the adminClient in line 35.

extensions/harvester/sakubeconfig.go
```diff
@@ -0,0 +1,98 @@
...
+
+	req, err := http.NewRequest("POST", fmt.Sprintf("https://%s/k8s/clusters/%s/v1/harvester/kubeconfig", adminClient.RancherConfig.Host, status.ClusterName), bytes.NewBuffer(bodyContent))
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+adminClient.RancherConfig.AdminToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	response, err := adminClient.Management.APIBaseClient.Ops.Client.Do(req)
```
anupama2501, line 80: should this be client instead of adminClient for any rbac related cases?

extensions/harvester/sakubeconfig.go
igomez06: Good question, can we always assume it's an admin client?

slickwarren: yeah sorry you guys are right, this can just be the client of the user who needs it.

extensions/harvester/sakubeconfig.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest("POST", fmt.Sprintf("https://%s/k8s/clusters/%s/v1/harvester/kubeconfig", client.RancherConfig.Host, status.ClusterName), bytes.NewBuffer(bodyContent))
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+client.RancherConfig.AdminToken)
```
anupama2501, line 72: also might need to change this. we have admin token here. if the client is not admin, the request might fail.

igomez06 (dismissed):
LGTM

extensions/harvester/sakubeconfig.go
anupama2501: Since, the adminClient was removed from the current version, the client we use will be a non admin client. In that case a non admin client would be using the admin token..

igomez06: So Caleb if we're keeping it accessible to standard non-admin clients then the clients bearer token should be used like this: 
client.Management.APIBaseClient.Opts.TokenKey try this out

anupama2501 (approved):

igomez06 (approved):
LGTM

PR #390, opened 2025-04-10
Use BackingNamespace for projects. Update pkg/apis

JonCrowther:
To unblock this issue https://github.com/rancher/rancher/pull/49797 we need to make shepherd aware of the new BackingNamespace field in projects. Newly created projects will use `BackingNamespace` as the namespace for resources needed in projects. The `AddProjectMember` function watches that namespace to ensure it gets created, which is what's failing in the integration tests.

I also ran `go generate` which updated a lot of files. Additionally, the go version needed to be bumped in order to use the latest rancher/rancher changes.


go.mod
```diff
@@ -15,10 +15,10 @@ replace (
 )
```
caliskanugur, line 15: I think CI fails due to compatibility with upstream packages cause of the replaces here.

Could you map these k8s packages to `v0.32.2` instead of `v0.31.1` and see if it helps?

caliskanugur (commented):
Hey Jonathan, thank you for bumping the packages as well! 

Only one thing related to CI, other than that LGTM

go.mod
```diff
@@ -1,24 +1,24 @@
...
 replace (
 	github.com/docker/distribution => github.com/docker/distribution v2.8.2+incompatible // rancher-machine requires a replace is set
-	k8s.io/api => k8s.io/api v0.31.1
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.1
-	k8s.io/apimachinery => k8s.io/apimachinery v0.31.1
-	k8s.io/apiserver => k8s.io/apiserver v0.31.1
-	k8s.io/cli-runtime => k8s.io/cli-runtime v0.31.1
-	k8s.io/client-go => k8s.io/client-go v0.31.1
+	k8s.io/api => k8s.io/api v0.32.2
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.31.2
```
igomez06, line 10: @JonCrowther is this the correct version?

caliskanugur (approved):
LGTM

PR #391, opened 2025-04-10
Undo changes to project namespace

JonCrowther:
I misdiagnosed the cause of the integration test failures in https://github.com/rancher/rancher/pull/49797. I thought that the issue was the namespace being watched, but it was actually the Condition not being satisfied. Reverting.


caliskanugur (approved):
LGTM

igomez06 (approved):
LGTM

PR #407, opened 2025-08-05
Updated pkgs to stay in sync with rancher/rancher.

igomez06:
## Problem
 Some version in the main version of shepherd were out sync with rancher/rancher
 
## Solution
 sync up rancher versions
 
_TODO_


rancher-max (approved):

go.mod
```diff
@@ -2,13 +2,13 @@ module github.com/rancher/shepherd
...
 	github.com/docker/distribution => github.com/docker/distribution v2.8.2+incompatible // rancher-machine requires a replace is set
-	k8s.io/api => k8s.io/api v0.32.2
-	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.32.2
-	k8s.io/apimachinery => k8s.io/apimachinery v0.32.2
+	k8s.io/api => k8s.io/api v0.33.2
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.33.2
+	k8s.io/apimachinery => k8s.io/apimachinery v0.33.2
 	k8s.io/apiserver => k8s.io/apiserver v0.32.2
 	k8s.io/cli-runtime => k8s.io/cli-runtime v0.32.2
 	k8s.io/client-go => k8s.io/client-go v0.32.2
```
Priyashetty17, line 14: Looks like a few instances of k8s dependencies were updated from 0.32.2 to 0.33.2. Shouldn’t we update all k8s.io dependencies to match what’s in rancher/rancher? F.E k8s.io/kubectl, k8s.io/api, k8s.io/apiextensions-apiserver etc

igomez06: good eye. I think updated the ones I missed.

Priyashetty17: 😬 just one more - k8s.io/pod-security-admission is still at 0.32.1

igomez06: I skipped the indirect imports but I got those too 😁

igomez06: Hi Priya so I changed it anyway,  but since this is an indirect import that version is dependent on whatever we are importing directly. In other words we're using the version in the direct imports. I changed it this time, but moving forward we don't necessarily want to change these especially if we're automating it, which is the plan. The packages need to makes sure they're updating their own go mods as well. Does that make sense?

Priyashetty17 (approved):
lgtm

rancher-max (approved):

PR #413, opened 2025-08-18
Update SSH functions to work with a non-default ssh user

susesgartner:
Rancher PR: https://github.com/rancher/tests/pull/195

The current implementation of various SSH functions utilizes the default SSH username (ubuntu) in rancher, which causes any other OS type to fail.

Fix:
* Reworked the sshKeys extension to pull username, IP and sshKey from the v1/cluster.x-k8s.io.machines api endpoint

### DIFF
--- a/extensions/sshkeys/downloadsshkeys.go
+++ b/extensions/sshkeys/downloadsshkeys.go
@@ -1,109 +1,101 @@
 package sshkeys
 
 import (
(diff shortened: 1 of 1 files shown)


igomez06 (approved):
LGTM can we get back ports as well Sam? Thanks!

Josh-Diamond (approved):

PR #427, opened 2025-09-16
[main] Extend RKE2/K3s restore snapshot time

markusewalker:
## Problem

In our recurring runs for this release testing cycle, it was noted that some snapshot tests were producing a red herring failure. Digging deeper, it is because the restore time has a very short window of two minutes. This needs far more of a buffer as it takes longer than two minutes at times.
 
## Solution

Update the timeout to 30 minutes.


Josh-Diamond (approved):

caliskanugur (approved):
LGTM

PR #438, opened 2025-09-17
[v2.12] Fixed norman version to match rancher/rancher.

igomez06:
## Problem
 The norman version was accidentally upgraded to 0.7.0, and it was causing the norman version to upgraded in rancher/rancher

 
## Solution
 Downgrade back to v0.6.1
 
_TODO_


caliskanugur (dismissed):
LGTM

PR #440, opened 2025-09-19
Update nodes.go to allow bogh IPv6 and IPv4 connections

markusewalker:
## Problem

As part of broader IPv6 testing, custom cluster provisioning needs to be tweaked a bit to allow this. Specifically, when SSH'ing into a node or copying files over to a node, `nodes.go` only allows IPv4.
 
## Solution

Created a small helper function that will determine if the IP address is IPv4 or IPv6, depending on if a `:` is identified in the address. Very small change.
 
## Testing

Jenkins jobs will be provided offline.


Josh-Diamond (approved):

igomez06 (approved):
LGTM

PR #446, opened 2025-10-03
Add subnetID parameter for EC2

markusewalker:
## Problem

The current `config.go` does not allow for choosing your own subnet. As part of broader IPv6 support, we need the option to specify an IPv6-only subnet as part of the networking.
 
## Solution

Added `AWSSubnetID` parameter.


igomez06 (approved):
LGTM

PR #549, opened 2026-04-10
Add feature flag extensions using wrangler context

Priyashetty17:
This PR adds new extensions to support updating (enabling/disabling), and checking the status of feature flags using the Wrangler context.
- **Feature Flags** (IsFeatureEnabled/UpdateFeatureFlag): Generic CRUD helpers for the Feature resource using Wrangler context.
- **GetClusterWranglerContext**: Generic helper to retrieve a cluster Wrangler context.
- **WaitForDeploymentActive**: Generic helper that waits for a Deployment to become active by polling until replicas are ready. I added this as an extension since wait-based helpers on resources seems to be acceptable as per the guidelines in "Actions vs Extensions" doc? It is useful for scenarios such as feature flag updates where Rancher restarts and we wait for it become active.

### DIFF
--- a/extensions/kubeapi/cluster/cluster.go
+++ b/extensions/kubeapi/cluster/cluster.go
@@ -1,0 +1,24 @@
+package cluster
+
+import (
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/pkg/wrangler"
+)
+
+const (
+	LocalCluster = "local"
+)
+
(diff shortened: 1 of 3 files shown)


Priyashetty17:
@dasarinaidu mentioned his PR depends on the helpers in this PR and will copy them into his PR to avoid waiting for this PR to be merged, so I’m closing this PR to avoid duplication.

PR #555, opened 2026-04-24
[Main] Add networking controller and ingress kubeapi extensions (wrangler-based)

Priyashetty17:
This PR adds networking controller support to Shepherd and introduces Wrangler-based **ingress** kubeapi helpers so tests can use Wrangler instead of dynamic client helpers.

Having a mix of dynamic client and Wrangler based helpers has caused confusion, especially for new test writers. This PR is part of a broader effort [here](https://github.com/rancher/qa-tasks/issues/2323) to refactor some of the kubeapi helpers to reduce that confusion without impacting any existing tests.

After this PR is mergeed, existing **ingress** helpers in rancher/tests/actions that use the dynamic client (not Steve) will be removed, and any tests (Example: https://github.com/rancher/tests/blob/main/validation/rbac/ingress/ingress_rbac_test.go) using them will be updated to use the new extensions.

The new CRUD helpers in this PR are added as Extensions rather than Actions because they are generic, reusable ingress CRUD and helpers without test-specific defaults or assertions based on the documentation here: https://confluence.suse.com/spaces/RANQA/pages/1505559041/Actions+vs+Extensions


igomez06 (approved):
LGTM let's get backports before merging please

PR #560, opened 2026-04-29
 Bump rancher/pkg/apis, cluster-api, and otel pins

dasarinaidu:
**Bump rancher/pkg/apis, cluster-api, and otel pins**
                                                                                                                                                                                                                                                              
  - github.com/rancher/rancher/pkg/apis to latest pseudo-version
    (exposes new fields like OIDCClientSpec.Scopes)                                                                                                                                                                                                           
  - sigs.k8s.io/cluster-api replace v1.8.3 -> v1.12.2
    (api/v1beta1 moved to api/core/v1beta1)                                                                                                                                                                                                                   
  - go.opentelemetry.io/otel pins v1.28.0 -> v1.38.0                                                                                                                                                                                                          
    (transitively required by the newer pkg/apis dep tree)
  - Update import path in actions/provisioning/verify.go                                                                                                                                                                                                      
  - Apply across root, actions, and interoperability submodules


.github/workflows/verify-changes.yaml
```diff
@@ -37,9 +37,9 @@ jobs:
         run: ./.github/scripts/build-packages.sh
 
       - name: Golangci Lint
-        uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
+        uses: golangci/golangci-lint-action@v8
         with:
           # Patch version isn't needed; https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#internals
-          version: v1.64.6
+          version: v2.5.0
```
Priyashetty17, line 43: Is this change intentional and why is it updated?

go.mod
```diff
@@ -30,31 +28,31 @@ require (
...
-	golang.org/x/crypto v0.43.0
+	golang.org/x/crypto v0.49.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.34.1
-	k8s.io/apimachinery v0.34.1
-	k8s.io/apiserver v0.34.1
+	k8s.io/api v0.35.4
+	k8s.io/apimachinery v0.35.4
+	k8s.io/apiserver v0.35.4
 	k8s.io/cli-runtime v0.34.1
```
Priyashetty17: I see some k8s versions are updated to 0.35.4, while others are still at 0.34.1. Could you ensure they are all updated?

.github/workflows/verify-changes.yaml
```diff
@@ -37,9 +37,9 @@ jobs:
         run: ./.github/scripts/build-packages.sh
 
       - name: Golangci Lint
-        uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # v6.5.2
+        uses: golangci/golangci-lint-action@v8
```
Priyashetty17: Is this change intentional and why is it updated? If this was intentional, you need to pin the version to a full-length commit ID instead of a tag.

go.mod
```diff
@@ -30,31 +28,31 @@ require (
...
-	github.com/rancher/lasso v0.2.5
-	github.com/rancher/norman v0.7.1
-	github.com/rancher/rancher/pkg/apis v0.0.0-20251107181200-efcbd5d43f4e
-	github.com/rancher/rke v1.8.0-rc.4
+	github.com/rancher/eks-operator v1.14.0-rc.5
+	github.com/rancher/fleet/pkg/apis v0.15.0-beta.4
+	github.com/rancher/gke-operator v1.14.0-rc.3
+	github.com/rancher/lasso v0.2.7
+	github.com/rancher/norman v0.9.1
+	github.com/rancher/rancher/pkg/apis v0.0.0-20260429165553-4ae6ee5ee410
```
Priyashetty17: The latest version appears to be `v0.0.0-20260428160942-f131ed782a24`

.github/workflows/verify-changes.yaml
dasarinaidu: Yes, intentional. The pkg/apis bump pulls in transitive deps (aks-operator, eks-operator, gke-operator, fleet, lasso, norman) that all require Go 1.25. The previously pinned golangci-lint v1.64.6 was built with Go 1.24 and can't parse Go 1.25 modules
CI was failing with the Go language version (go1.24) used to build golangci-lint is lower than the targeted Go version (1.25.0). v1.x has no Go 1.25 release (the project moved to v2 series). v2.5.0 is built with Go 1.25.1.

go.mod
dasarinaidu: Good catch — fixed. All k8s.io/* replace directives now aligned at v0.35.4 to match the transitive bump pulled in by the new pkg/apis. (Was a mix of v0.34.1 and v0.35.4 because the replace block hadn't been updated.

.github/workflows/verify-changes.yaml
dasarinaidu: Yes, intentional and tied to comment #1: action v6.5.2 was pinned for golangci-lint v1.x; v2.x lint needs action v7+. Pinned to commit SHA per shepherd convention: golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.

go.mod
dasarinaidu: Bumped to v0.0.0-20260430130736-3f1ff49a72cb — that's today's latest from the Go proxy

.github/workflows/verify-changes.yaml
caliskanugur: We want parity between shepherd and rancher/rancher, this PR already covers that parity, could you take a look at it? https://github.com/rancher/shepherd/pull/536

caliskanugur (commented):
Some maybe most of these changes are covered in https://github.com/rancher/shepherd/pull/536

If this PR is not an immediate blocker, could you please give that PR a review for the same bits that this one is changing

.github/workflows/verify-changes.yaml
dasarinaidu: Thanks for flagging #536 

 A few things to consider:
 #560 is a focused 5-file bump intentionally kept small to unblock the OIDC PR (#547), which Priya's already reviewed and approved pending the typed-wrangler rewrite that depends on v3.OIDCClientSpec.Scopes. #536 is a larger 408-file change (RKE1 removal) that's been in cross-team review for breaking changes since March, currently using a fork pin for pkg/apis (bhartigautam156/rancher/pkg/apis).

Would it make sense to land #560 as the focused dep bump (no behavior changes outside the bump itself) so OIDC can move forward, then handle the RKE1 removal on its own timeline in #536?

PR #567, opened 2026-05-20
Add skipSchemaValidation to catalog APIs

alemorcuq:
## Problem
Adding the SkipValidationSchema field in rancher/rancher#55004 but can't be tested because it is not in Shepherd.
 
## Solution
Update the catalog API with the new fields.
 
## Testing
It adds a new field that is omited by default, so there shouldn't be any issues.

It adds a new field that is omited by default, so there shouldn't be any issues.


hamistao (approved):
LGTM

lscalabrini01 (approved):

Issue #569, opened 2026-05-20
`GetRegistrationToken` panics on empty list and times out too quickly at scale/on slow environments


Oats87:
# Problem

 `GetRegistrationToken` in `extensions/tokenregistration/tokenregistration.go` has reliability issues that surface in large-scale Rancher test environments:

  1. Panic on empty list. After the poll loop, the function blindly indexes clusterRegistrationTokens[0] without checking length. If the poll exits with an empty slice (e.g., on timeout, or if the API returns an empty page), callers get a runtime error: index out of range instead of a clean error.
  2. Returned token may not be usable. A ClusterRegistrationToken can appear in the list before Rancher's controllers have populated its Token and ManifestURL fields. Callers that immediately consume the returned token to register a node will fail with confusing errors.
  3. Timeout too aggressive. The current 2-minute ceiling assumes a low-load environment. Measured token-creation lag at 500 downstream clusters is p50=9s, p95=78s, p99=141s for the default token and p50=113s, p95=249s, max=315s for the system token. At 999 clusters the tail extends further. The handler is chained behind ~15 other Mgmt.Cluster() handlers, so lag scales with cluster count.

## Proposed change

  - Replace deprecated kwait.Poll with kwait.PollUntilContextTimeout.
  - Use ListAll instead of List (consistent with the rest of `extensions/clusters`).
  - Filter the returned slice for the first token whose Token AND ManifestURL are both populated; treat unpopulated entries as "still waiting" rather than success.
  - Return a wrapped error on timeout instead of indexing [0] of an empty slice.
  - Raise the poll timeout to 15 minutes based on the measurements above. Tokens that haven't appeared by then almost always indicate a stuck cluster, not lag.
  - Emit a `logrus.Warnf` at a quiet threshold (~60s) so short happy-path runs stay quiet while long stalls become visible.

## Caller impact

This is a behavior change — today's caller silently gets a zero-value token (or panics on empty list); the new version returns a real error. Callers that ignore the error or nil-check the return will regress.

git-ival:
Closed by #571

PR #571, opened 2026-05-20
tokenregistration: poll with backoff and return error on timeout

Oats87:
## Issue: 
https://github.com/rancher/shepherd/issues/569
 
## Problem
GetRegistrationToken indexed [0] of the polled-token slice without a length check, causing a panic when the list was empty after timeout. It also returned tokens whose Token/ManifestURL had not yet been populated by Rancher's controllers, so callers using the returned token to register nodes failed with confusing errors. The 2-minute ceiling was too short at scale: p99 token-creation lag is 141s at 500 downstream clusters and extends further at 999.

## Solution
Switch to PollUntilContextTimeout, use ListAll, filter for tokens whose Token AND ManifestURL are both populated, return a wrapped error on timeout, and raise the ceiling to 15 minutes. Add a quiet-log threshold so short happy-path runs stay quiet while long stalls become visible.

GetRegistrationToken now returns a real error instead of a zero-value token; callers that previously ignored err will need to handle it.

### DIFF
--- a/extensions/tokenregistration/tokenregistration.go
+++ b/extensions/tokenregistration/tokenregistration.go
@@ -8,42 +8,20 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/shepherd/clients/rancher"
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	"github.com/rancher/shepherd/extensions/defaults"
 	"github.com/sirupsen/logrus"
 	kwait "k8s.io/apimachinery/pkg/util/wait"
 )
 
-const (
-	// registrationTokenPollInterval is the interval between polls for a populated
-	// ClusterRegistrationToken.
-	registrationTokenPollInterval = 2 * time.Second
-
-	// registrationTokenPollTimeout is the per-cluster ceiling for waiting on
-	// Rancher to populate ClusterRegistrationToken.Token and .ManifestURL.
-	//
-	// At ~500 downstream clusters we measured token-creation lag of
-	// p50=9s, p95=78s, p99=141s for the default token and p50=113s, p95=249s,
-	// max=315s for the system token. The leader-side cluster-watch handler is
-	// chained behind ~15 other Mgmt.Cluster() handlers, so lag scales with N
-	// — at 999 clusters the tail extends well past the 500-cluster numbers,
-	// so we give 15 minutes of headroom. Tokens that haven't appeared by then
-	// almost always indicate a stuck cluster, not lag.
-	registrationTokenPollTimeout = 15 * time.Minute
-
-	// quietLogThreshold suppresses the "no tokens listed yet" log line until
-	// the poll has been running long enough that absence is actually notable.
-	quietLogThreshold = 60 * time.Second
-)
-
-// GetRegistrationToken polls Rancher for a ClusterRegistrationToken belonging
-// to clusterID and returns the first one whose Token and ManifestURL are both
-// populated. An empty list (or one whose entries have not yet been filled in
-// by Rancher's controllers) is expected during the first ~10-170s of a
-// cluster's life and is not treated as fatal until the overall timeout.
+// GetRegistrationToken polls Rancher for a ClusterRegistrationToken belonging to clusterID
+// and returns the first one whose Token and ManifestURL are both populated, or an error on timeout.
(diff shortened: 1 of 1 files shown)


extensions/tokenregistration/tokenregistration.go
```diff
@@ -1,39 +1,88 @@
...
+	// quietLogThreshold suppresses the "no tokens listed yet" log line until
+	// the poll has been running long enough that absence is actually notable.
+	quietLogThreshold = 60 * time.Second
+)
+
+// GetRegistrationToken polls Rancher for a ClusterRegistrationToken belonging
+// to clusterID and returns the first one whose Token and ManifestURL are both
+// populated. An empty list (or one whose entries have not yet been filled in
+// by Rancher's controllers) is expected during the first ~10-170s of a
+// cluster's life and is not treated as fatal until the overall timeout.
```
Priyashetty17: nit, instead of defining new constants,  you could consider using the timeout constants from https://github.com/rancher/shepherd/blob/main/extensions/defaults/defaults.go

extensions/tokenregistration/tokenregistration.go
```diff
@@ -1,39 +1,88 @@
...
-
 		if err != nil {
-			return false, err
+			logrus.Errorf("[%s] error while listing cluster registration tokens: %v", clusterID, err)
+			return false, nil
 		}
 
 		if len(collection.Data) == 0 {
-			return false, err
+			if time.Since(start) >= quietLogThreshold {
```
Priyashetty17: Is the quietLogThreshold logic needed? Once `time.Since(start) >= quietLogThreshold` becomes true, this warning will be logged on every poll iteration (ie..,every 2 seconds) since the last log time is not tracked separately

extensions/tokenregistration/tokenregistration.go
```diff
@@ -1,39 +1,88 @@
...
+	registrationTokenPollInterval = 2 * time.Second
+
+	// registrationTokenPollTimeout is the per-cluster ceiling for waiting on
+	// Rancher to populate ClusterRegistrationToken.Token and .ManifestURL.
+	//
+	// At ~500 downstream clusters we measured token-creation lag of
+	// p50=9s, p95=78s, p99=141s for the default token and p50=113s, p95=249s,
+	// max=315s for the system token. The leader-side cluster-watch handler is
+	// chained behind ~15 other Mgmt.Cluster() handlers, so lag scales with N
+	// — at 999 clusters the tail extends well past the 500-cluster numbers,
```
slickwarren: Imo these comments aren't necessary or could be reduced to 1 line. But we normally don't have to justify timeout settings

extensions/tokenregistration/tokenregistration.go
```diff
@@ -1,39 +1,88 @@
...
 		}
 
-		clusterRegistrationTokens = collection.Data
-		return true, nil
+		if time.Since(start) >= quietLogThreshold {
+			logrus.Warnf("[%s] %d cluster registration token(s) listed but none populated after %s; still waiting", clusterID, len(collection.Data), time.Since(start).Round(time.Second))
+		} else {
+			logrus.Debugf("[%s] %d cluster registration token(s) listed but Token/ManifestURL not yet populated", clusterID, len(collection.Data))
+		}
+		return false, nil
```
Priyashetty17: Same comment as above, when the condition becomes true, this warning will be logged every 2 seconds because the last log time is not tracked separately

slickwarren (dismissed):
Just a nit about comments

extensions/tokenregistration/tokenregistration.go
Oats87: Using a `timeout` for a `threshold` is kind of a misnomer; however, I've gone ahead and changed over to using them for the log thresholds.

extensions/tokenregistration/tokenregistration.go
Oats87: I went ahead and changed the logic to a `log every N` instead doing log suppression-then-spam

extensions/tokenregistration/tokenregistration.go
Oats87: comments have been truncated

Priyashetty17 (approved):
lgtm

slickwarren (approved):

PR #584, opened 2026-06-09
[main] Cleanup related fixes for tests

Priyashetty17:
We have been seeing recurring resource-cleanup failures at test teardown.  Upon triaging, this required two changes. 
- **Extensions** - clean up with an admin client (not the creating client): 
    - **Problem**: RBAC tests create resources as a specific user. The controller auto-registers a cleanup that deletes the resource using the same client that created it. But create permission does not imply delete permission, so at teardown the user-scoped delete fails:
`Cleanup error: unable to delete (/, Kind=) c-m-kb2r4jjq/auto-testproject-gkjjc:
projects.management.cattle.io "auto-testproject-gkjjc" is forbidden:
		User "u-8wbx2" cannot delete resource "projects" in API group "management.cattle.io"`

    - **Change**: Explicitly register an admin-scoped cleanup
    
- **pkg/wrangler/pkg/generic/controller.go** - fix auto-cleanup for GenerateName resources  
    - **Problem**: Controller.Create captured the name/namespace for its auto-cleanup from the input object. For resources created with metav1.GenerateName, the input object's Name is empty, the real name is only assigned on creation. The cleanup therefore issued a delete with an empty name:
`		Cleanup error: unable to delete (/, Kind=) c-m-kb2r4jjq/: resource name may not be empty`
    - **Change**: Reordered Create so the object is created first, then the cleanup is registered. Auto-registered cleanup now targets the correct resource for both explicitly-named and GenerateName resources, eliminating the "resource name may not be empty" teardown errors.

**QA Task**: https://github.com/rancher/qa-tasks/issues/2323

### DIFF
--- a/extensions/kubeapi/configmaps/create.go
+++ b/extensions/kubeapi/configmaps/create.go
@@ -20,7 +20,12 @@ func CreateConfigMapWithTemplate(client *rancher.Client, clusterID string, confi
 
 	if client.Session != nil {
 		client.Session.RegisterCleanupFunc(func() error {
-			return DeleteConfigMap(client, clusterID, createdConfigMap.Namespace, createdConfigMap.Name, true)
+			adminClient, err := rancher.NewClient(client.RancherConfig.AdminToken, client.Session)
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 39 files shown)


igomez06 (approved):
LGTM

dasarinaidu (approved):

PR #589, opened 2026-07-13
Improved error on missing config key

hamistao:
This contains:
- Improved error on missing config key
- Creation of `UpdateSecretWithTemplate`
- Olhar small improvements

Also may contain changes from https://github.com/rancher/shepherd/pull/513 as it was missing from a few branches,

### DIFF
--- a/extensions/kubeapi/secrets/create.go
+++ b/extensions/kubeapi/secrets/create.go
@@ -33,3 +33,13 @@ func CreateSecretWithTemplate(client *rancher.Client, clusterID string, secretTe
 
 	return createdSecret, nil
 }
+
+func UpdateSecretWithTemplate(client *rancher.Client, clusterID string, secretTemplate *corev1.Secret) error {
+	clusterContext, err := extclusterapi.GetClusterWranglerContext(client, clusterID)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster context: %w", err)
+	}
+
+	_, err = clusterContext.Core.Secret().Update(secretTemplate)
+	return err
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,6 +19,7 @@ func LoadConfig(key string, config interface{}) {
 	configPath := os.Getenv(ConfigEnvironmentKey)
 
 	if configPath == "" {
+		logrus.Infof("No config file path provided via %s env var", ConfigEnvironmentKey)
 		yaml.Unmarshal([]byte("{}"), config)
 		return
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/creasty/defaults"
+	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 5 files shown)


extensions/kubeapi/secrets/create.go
```diff
@@ -33,3 +33,13 @@ func CreateSecretWithTemplate(client *rancher.Client, clusterID string, secretTe
...
 	return createdSecret, nil
 }
+
+func UpdateSecretWithTemplate(client *rancher.Client, clusterID string, secretTemplate *corev1.Secret) error {
+	clusterContext, err := extclusterapi.GetClusterWranglerContext(client, clusterID)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster context: %w", err)
+	}
+
+	_, err = clusterContext.Core.Secret().Update(secretTemplate)
```
caliskanugur, line 43: Would this function work in this scenario;

1. Instantiate/create a secret object
2. Update secret using (1)
3. Update secret using (1)

If working for this case ignore this comment please.

I believe third might fail due to it now needs to send updated secret object. If so, should this function return the updated object?

hamistao:
Backports for this (not all changes apply to all the branches):

2.14: https://github.com/rancher/shepherd/pull/591
2.13: https://github.com/rancher/shepherd/pull/592
2.12: https://github.com/rancher/shepherd/pull/597
2.11: https://github.com/rancher/shepherd/pull/596

@igomez06

pkg/config/config.go
```diff
@@ -33,7 +35,10 @@ func LoadConfig(key string, config interface{}) {
 		panic(err)
 	}
 
-	scoped := all[key]
+	scoped, ok := all[key]
```
caliskanugur, line 38: If scoped is `nil`, I think marshal -> unmarshal down the line may instantiate a zero value data type/struct for the key. 

Panicking over it might change the flow of things like my question below, since it deliberately looks for a key. 

Would this be a problem in cases where configuration operations happen in a pipeline? Adding @susesgartner to the PR, he might have better insights on this change

caliskanugur (commented):
Looks good, just have couple questions out of curiosity

extensions/kubeapi/secrets/create.go
hamistao: Great question! Though I haven't tired this, I believe this should work since you are using the same secret object/template for all operations then it will always represent the current state of the secret after any operation. So even if you returned the object it would be just a copy of the one you passed in if the update was successful.

pkg/config/config.go
hamistao: Aren't we already panicking over it, just with a less useful message?

lscalabrini01 (dismissed):

pkg/config/config.go
igomez06: So we panic if there is a yaml/json misconfiguration. What you're adding is a panic if a key doesn't exist. Is that new behavior we want to add?

hamistao: Sorry I am having trouble understanding this. The way I see it the only thing I am adding is a different error message on panic.

Currently, if the key doesn't exist, `scoped` will be nil and then we will panic soon after with a nil pointer deference error. The difference is that now you have a more informative message. Am I missing something?

And if there is a case where we would like to avoid panicking on a missing config key we just use `recover()`, as is already done in places such as https://github.com/rancher/tests/blob/d496766c4f3f413de9ac2ab99104c62200e25fa2/interoperability/longhorn/longhorn.go#L20.

hamistao: If you prefer I can just add a logrus.Infof instead of panicking.

hamistao: We decided here to just log the missing key message

caliskanugur (approved):
LGTM

igomez06 (approved):
LGTM

PR #598, opened 2026-07-20
Surface kubectl job failures instead of masking them

floatingman:
## Problem
`kubectl.Command()` masked the real cause of helper-Job failures. It only re-surfaced non-timeout `net.Error`s from `CreateJobAndRunKubectlCommands`; the watch timeout from `wait.WatchWait` (a plain error) and other non-`net.Error` failures were silently discarded. It then proceeded to fetch the job pod's logs, and when no pod matched, `GetPodLogs("")` failed with a misleading **`resource name may not be empty`** — hiding the actual root cause (e.g. an unpullable `shell-image` in an airgap cluster).

## Fix
- Capture the job error as `jobErr` instead of swallowing it, while **preserving** the best-effort log fetch (pod logs usually explain the failure).
- Guard the empty-`podName` case with a clear error that includes `jobErr`.
- Augment the log-stream failure with `jobErr` when the job also failed.
- Behavior is unchanged when logs are retrievable — still returns `(logs, nil)`.

## Verification
Reproduced the failure condition (unpullable `shell-image`) against a live cluster via `rancher/tests`:
- **Before:** `error streaming pod logs for pod kube-system/: resource name may not be empty`
- **After:** `kubectl job kubectl-hrglgn failed (job error: error with watch connection…); … container … is waiting to start: image can't be pulled`

`gofmt` clean; `go vet` / `go build ./extensions/kubectl/` pass.

### DIFF
--- a/extensions/kubectl/command.go
+++ b/extensions/kubectl/command.go
@@ -3,7 +3,6 @@ package kubectl
 import (
 	"errors"
 	"fmt"
-	"net"
 	"strings"
 
 	namegen "github.com/rancher/shepherd/pkg/namegenerator"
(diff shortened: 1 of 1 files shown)


floatingman:
## 🔗 Related PRs — full fix chain for the airgap `resource name may not be empty` failure

**This PR (shepherd) is the error-reporting improvement** — it is independent and not strictly required for the fix, but it stops masking the real cause.

`WorkloadTestSuite/TestDeployments/WorkloadUpgradeTest` (and any other path using `kubectl.Command`) fails in airgap because the rancher-shell helper Job can't pull its image. Three changes address different layers:

- [ ] **rancher/shepherd#598** ← this — stop masking the real kubectl-Job failure. Error now reports `"image can't be pulled"` instead of the misleading `"resource name may not be empty"`.
- [ ] **rancher/qa-infra-automation#123** — (root cause) `airgap_rancher_helm_deploy` defined `rancher_system_default_registry` but never passed it to the chart; now sets `global.cattle.systemDefaultRegistry` at install time so `shell-image` resolves to the private registry.
- [ ] **rancher/tests#771** — supplies the value: renders `rancher_system_default_registry: "${PRIVATE_REGISTRY_URL}"` in the airgap-rke2-tests group_vars.

**Dependency:** tests#771 sets the variable; qa-infra-automation#123 makes the role consume it. qa-infra-automation#123 is the root-cause fix; tests#771 supplies the value. This shepherd PR is independent.

**Also required (ops):** mirror `rancher/shell:<tag>` to `<registry>/rancher/shell:<tag>` — the containerd `proxycache`/`quaycache` mirrors do not cover this verbatim reference.

hamistao (approved):

lscalabrini01 (approved):

PR #705, opened 2026-09-03
Add Keycloak SAML provider and Keycloak admin API clients

dasarinaidu:
## Problem
Shepherd has no client for Keycloak and no client for Rancher's SAML auth providers, so the Keycloak
SAML provider cannot be automated the way OpenLDAP and Active Directory already are.

## Solution
- `clients/keycloak` — admin API client for realms, clients, protocol mappers, groups and users.
- `clients/rancher/auth/saml` — enables, disables and updates a Rancher SAML auth config, and
  generates the service provider key pair that Rancher requires but does not create itself.
- `extensions/auth/saml` — raw HTTP helpers for driving Rancher's SAML endpoints and for parsing the
  assertions an identity provider issues.
- `clients/rancher`: adds `AsSAMLUser`. SAML cannot go through `AsAuthUser`, because the login action
  answers with a redirect to the identity provider rather than a token.
- `extensions/defaults`: adds `OneSecondTimeout`.

The provider client is parameterised by a `Provider` value rather than hardcoded to Keycloak, so Okta
and the other SAML providers can be added by naming one and pointing it at its own config key.


extensions/auth/saml/diagnose.go
```diff
@@ -0,0 +1,65 @@
...
+	"regexp"
+	"strings"
+	"time"
+)
+
+const IdPResponseDumpDirEnvVar = "SAML_IDP_DUMP_DIR"
+
+var pageTitlePattern = regexp.MustCompile(`(?is)<title[^>]*>(.*?)</title>`)
+
+func describeIdPResponse(response *http.Response, body string) string {
```
joesims22, line 18: Nit, could you add a comment for this function, i.e. Helper func to diagnose/debug SAML IdP response.

joesims22 (commented):
One nit, otherwise lgtm.

joesims22 (approved):
lgtm

igomez06 (approved):
LGTM

